PR #1, opened 2014-03-17
Stripping out project and derived files

anjackson:
I believe it's common practice to avoid checking in project files (e.g. Eclipse's .project/.classpath/.settings) and derived or compiled files (e.g. content on /target/ for Maven projects). The former is because the project files tend not to be portable, and the latter is just because it's unnecessary, I think.

This pull request removes those files from the repo, and adds a .gitignore file to prevent then being added again in the future.

Accepting this will cause conflicts with or wipe out the existing project files, so they will have to be re-generated by each developer.


gmcgath:
Eclipse seemed to be trying to gitignore the derived files, so any that were left in were unintentional.

With the project file, a search turns up a fair amount of debate on whether they're portable or not. In practice, I think Eclipse users use the Maven import function to get their projects, which should make the .project file unnecessary. In any case, this is an area where I don't have much experience to go on, so I'll defer to you on it.

PR #3, opened 2014-03-27
[unit tests] add some unit tests

willp-bl:
Not all pass right now, happy for tests to be changed/deleted if not applicable.


willp-bl:
NOTE: the files in the links are not from this mavenised Jhove so can't be directly pulled.

I have some black-box testing style unit tests; however, they could do with some work as they feel a bit brittle.  See: https://github.com/willp-bl/jhove-mavenised/tree/jhove1xunittest/src/test/java/jhove1/output/test

They rely on a JhoveAPI class I added: https://github.com/willp-bl/jhove-mavenised/blob/jhove1xunittest/src/main/java/jhove1x/JhoveAPI.java

It's be great to see something like the JhoveAPI class added.

gmcgath:
I like unit tests, but perhaps they should be brought to maturity in your fork and then merged.

willp-bl:
Do you mean the ones in this pull request, or the ones I mention in the comment? (or both).  If it's the ones in this pull request then specific comments would be useful.

PR #6, opened 2014-03-28
Build two executable jars - console app and GUI app (double click to launch GUI app)

willp-bl:


gmcgath:
I've created a new branch called pull_staging, for pull requests that look reasonable but are non-trivial. This pull request has been merged into it (right now that's the only difference from master). I merged two other pull requests, which were more obviously right, directly.

Issue #9, opened 2014-08-21
JHOVE Incorrectly reading beyond RIFF 'data' Chunk ID and calling it invalid...


ross-spencer:
I have received a 2GB wav file that I'm having difficulty validating in JHOVE. The tool tells me that I have an invalid character within a CHUNK ID. 

Analyzing the file, however, and it seems that JHOVE is reading beyond the CHUNK ID and returning an invalid result. 

```
52 49 46 46 
Chunk ID: 'RIFF'

F8 DE A0 84 
Chunk Size: ~2GB

57 41 56 45 
Format: 'WAVE'

66 6D 74 20 
Sub Chunk 1 ID: 'fmt'

10 00 00 00 
Sub Chunk 1 Size: 16

01 00 
Audio Format: WAVE_FORMAT_PCM

01 00 
Number of Channels: 1

00 77 01 00 
Sample Rate: 96000

00 65 04 00 
Byte Rate: 288,000

03 00 
Block Align: 3

18 00 
Bits per sample: 24-bits

64 61 74 61 
Sub Chunk 2 ID: 'data'

80 C6 A0 84 
Sub Chunk 2 Size: ~1.6GB

A7 *05 00* 70 04 00 6E F6 FF E9 FC FF F7 F4 FF B5 24 00
... data / payload ...
```

The error message seems to be returned from this part of the code: 

https://github.com/gmcgath/jhove/blob/0dc774d98efa8c7581fe1602c3f6e713f499201d/src/main/java/edu/harvard/hul/ois/jhove/module/iff/ChunkHeader.java#L53

The byte causing the first issue is 0x05 at offset 46, I've starred offset 46 and 47. See also the screenshot. 

The screenshot has been generated by looking at the following snippet from the 2GB file:

```
52 49 46 46 F8 DE A0 84 57 41 56 45 66 6D 74 20 10 00 00 00 01 00 01 00 00
77 01 00 00 65 04 00 03 00 18 00 64 61 74 61 80 C6 A0 84 A7 05 00 70 04 00 
6E F6 FF E9 FC FF F7 F4 FF B5 24 00 F2 FC FF 88 FC FF 2C E8 FF 1B 08 00 74 
03 00 26 EE FF 20 F6 FF 86 F6 FF 33 01 00 5F F3 FF C0 FC FF 47
```

[...]

PR #10, opened 2014-08-21
Support WAV chunks > Integer.MAX_INT

anjackson:
This pull request addresses #9, including that truncated WAV as a test file. It seems the chunk size was assumed to be less that Integer.MAX_INT, which is problematic because the size can be up to that of an unsigned integer (which is that maximum chunk size for WAV I think - I wonder what happens if the WAV is larger still?).

I added the sample3.wav and created a unit test that reproduced the error, and then tracked down enough (int) operations to make it work again. This does not mean this issue will be resolved for other format modules, although there were modifications in a few modules to account for ModuleBase.skipBytes now returning a long rather than an int.


carlwilson:
Refactored merge against new Maven project structure: openpreserve/jhove#37, this will be closed when  the new PR is merged if satisfactory.

carlwilson:
I believe that this is now closed by #37  that took into account location changes following the Maven refactoring.

PR #14, opened 2014-12-10
Only creates one jhove.jar that generates jhove java source from jhove.xsd

jansoren:
- Removed unnecessary creation of gui and cmd jar, since they both were identical
- Now generating jhove java source for unmarshalling of the jhove xml result
- Made a release version 1.12.0
- Made a snapshot version 1.13.0-SNAPSHOT

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ pom.xml.tag
 pom.xml.releaseBackup
 pom.xml.next
 release.properties
+/.idea/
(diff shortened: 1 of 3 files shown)


ravn:
There is a new release 1.12 in your commits which is not reflected on the title of the pull request.  Perhaps you could cherry pick the needed commits for the functionality you on a new branch and then do a pull request on that?

PR #47, opened 2015-09-08
Added appassembler plugin to create working target/appassembler/bin/j…

ravn:
…hove during build.

We would like to use jhove as a library for our own application, but it is designed to be used from the command line only.  In order to enclose it in our build (as a git submodule providing a maven module) I have added appassembler-maven-plugin support in order to create a "jhove" binary in jhove/jhove-apps/target/appassembler during the maven build we can call from integration tests etc.

Due to the design of jhove a configuration file _must_ be present in the file system.  I have added a copy of the jhove.conf file found in jhove-installer and set joveHome to "unused", and let the unix wrapper know about it.

### DIFF
--- a/jhove-apps/pom.xml
+++ b/jhove-apps/pom.xml
@@ -16,6 +16,51 @@
 
   <build>
     <plugins>
+      <!-- create the tarball layout -->
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>appassembler-maven-plugin</artifactId>
+        <version>1.10</version>
+        <configuration>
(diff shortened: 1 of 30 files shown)


carlwilson:
Sorry @ravn is this still necessary? I'm doing a thorough JHOVE spring clean and have been looking at this. I'm about to try to do away with the config file altogether (almost) and overhaul the build structure. If you still want this in some form I'm willing to throw in a profile build for Maven. If I've not heard anything by end of 2022 I'll probably close this but it'll be easier to run/test JHOVE by then also.

ravn:
Thank you for coming back to me on this.   I am no longer with kb.dk, so I have no personal interest in this any longer.   Thanks again.

PR #54, opened 2015-11-05
Improve batch script

JayXon:
Use %~dp0 to get install path, no need for substitution.
Use %\* to get command line arguments, no need for loop.


carlwilson:
Included in #245

PR #57, opened 2015-12-11
Corrected usage examples + added more visible link to jhove homepage

carlwilson:


carlwilson:
Pointed existing #56 at integration branch for now as an update is nigh.

PR #92, opened 2016-05-26
Removing final keyword from parsing method in jhove modules

johnscancella:
this way anyone can easily extend a module that already exists.

### DIFF
--- a/jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/AsciiModule.java
+++ b/jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/AsciiModule.java
@@ -145,7 +145,7 @@ public class AsciiModule extends ModuleBase {
      * RepInfo.
      */
     @Override
-    public final int parse(InputStream stream, RepInfo info, int parseIndex)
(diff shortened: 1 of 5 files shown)


carlwilson:
Hi John,

I looked at the use of final pretty closely as it sounded like something I might do. I'm almost certain it's not my code but there is a logic behind using final unless the class itself has been designed with inheritance in mind. The thinking is that most classes aren't designed to be extended and good/reliable design here can be time-consuming. By making the method final the developer indicates that it's possibly unsafe/unreliable to replace the method or extend the class. OR It could just be a mistake.

gmcgath:
It's been a long time, but I think I used "final" extensively, on the assumption that modules shouldn't be subclassed. The problem is largely with the package structure. The primary class of each module is in edu.harvard.hul.ois.jhove.module, and all the rest of the code is in edu.harvard.hul.ois.jhove.module.[modulename]. This was a poor design decision that forced a lot of the code in the primary class to be public. Making it final was a way to put some limits on it. (Or at least that's my best recollection.) Any changes to allow subclassing ought to be done with some thought, rather than just stripping "final" off everything.

PR #97, opened 2016-07-19
Addressing issue #93: CrossRefStream incorrectly assumes /Index value is a 2 element array

paulmer:
array containing an arbitrary set of starting object number and object
count pairs instead of a single pair.

Calculate number of objects using final start/count pair.  (Actual number
will be less than or equal to this value.)

Modified reading loop to read only objects in the index ranges.

Modified PdfModule to test object number against stream's index instead of
a range from 0 to the last object number.

Attempt to conform to PDF specification
http://wwwimages.adobe.com/content/dam/Adobe/en/devnet/pdf/pdfs/pdf_reference_archives/PDFReference15_v6.pdf, page 83 and PDF versions 1.6 and 1.7.

### DIFF
--- a/jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/CrossRefStream.java
+++ b/jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/CrossRefStream.java
@@ -25,12 +25,6 @@ import java.util.*;
  *
  */
 public class CrossRefStream {
-    
-	private class index_range {
-		public int start;
-		public int len;
-	};
-
     private PdfStream _xstrm;   // The underlying Stream object.
     private PdfDictionary _dict;
     private int _size;
@@ -51,7 +45,15 @@ public class CrossRefStream {
     private int _objNum;
     private int _objField1;
     private int _objField2;
-    
+
+	/** Range elements of the _index array:  
+		Starting object and number of objects.
+	*/
+	private class index_range {
+		public int start;
+		public int len;
+	};
+
     /**
      * Constructor.
      * 
--- a/jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/CrossRefStream.java
+++ b/jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/CrossRefStream.java
@@ -26,14 +26,23 @@ import java.util.*;
  */
 public class CrossRefStream {
     
+	private class index_range {
+		public int start;
+		public int len;
+	};
+
     private PdfStream _xstrm;   // The underlying Stream object.
     private PdfDictionary _dict;
     private int _size;
-    private int[] _index;
+	private int _index_size;
+    private index_range[] _index;
     private int[] _fieldSizes;
     private int _freeCount;
     private Filter[] _filters;
     private int _entriesRead;
+	private int _read_range;
+	private int _read_index;
+
     private int _bytesPerEntry;
     private long _prevXref;    // byte offset to previous xref stream, if any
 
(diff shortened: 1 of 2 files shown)


smmorrissey:
Sheila +1

pmay:
+1

asciim0:
+1

jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/CrossRefStream.java
```diff
@@ -25,15 +25,18 @@
  *
  */
 public class CrossRefStream {
-    
     private PdfStream _xstrm;   // The underlying Stream object.
     private PdfDictionary _dict;
     private int _size;
-    private int[] _index;
+	private int _index_size;
```
david-russo, line 31: `_index_size` appears to be unnecessary since the same value can be retrieved from `_index.length`.

jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/CrossRefStream.java
```diff
@@ -42,7 +45,15 @@
     private int _objNum;
     private int _objField1;
     private int _objField2;
-    
+
+	/** Range elements of the _index array:  
+		Starting object and number of objects.
+	*/
+	private class index_range {
```
david-russo, line 52: Similarly, all class names should be camel-case with an initial capital letter, i.e. `IndexRange`.

jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/CrossRefStream.java
```diff
@@ -262,8 +294,16 @@ public int getFreeCount ()
     /** Returns the total object count. */
     public int getNumObjects ()
     {
-        return _index[0] + _index[1];
+        return _index[_index_size - 1].start + _index[_index_size - 1].len;
```
david-russo, line 297: Like the original method, this seems like a convoluted way to calculate the value already stored in `_size`. Is there a reason we can't simply return that?

jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/CrossRefStream.java
```diff
@@ -25,15 +25,18 @@
...
     private PdfStream _xstrm;   // The underlying Stream object.
     private PdfDictionary _dict;
     private int _size;
-    private int[] _index;
+	private int _index_size;
+    private index_range[] _index;
     private int[] _fieldSizes;
     private int _freeCount;
     private Filter[] _filters;
     private int _entriesRead;
```
david-russo, line 36: With the addition of `_read_index` below, `_entriesRead` is no longer in use and can be safely removed.

jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/CrossRefStream.java
```diff
@@ -25,15 +25,18 @@
...
+	private int _read_range;
+	private int _read_index;
```
david-russo, line 38: Try and match any surrounding whitespace conventions. In this case 4 spaces instead of a tab. JHOVE's source code is currently a mess with indentation, but we should avoid adding to the confusion.

Also, the JHOVE (and [Java](http://www.oracle.com/technetwork/java/codeconvtoc-136057.html)) style convention for common variable names is to have them camel-cased with lowercase initial letters, e.g. `_readRange`, such as in the surrounding code.

jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/CrossRefStream.java
```diff
@@ -262,8 +294,16 @@ public int getFreeCount ()
     /** Returns the total object count. */
     public int getNumObjects ()
     {
-        return _index[0] + _index[1];
+        return _index[_index_size - 1].start + _index[_index_size - 1].len;
     }
+
+	public boolean isValidObject(int objNum) {
```
david-russo, line 300: Please add JavaDocs for any new methods.

david-russo (changes requested):
Thanks for the patch Paul, and good catch on the discrepancy with the spec. The code is largely fine and only requires some minor clean-up.

carlwilson:
Closed by #188

PR #118, opened 2016-09-09
#issue70 WaveModule->FormatChunk: ArrayIndexOutOfBoundsException

RogerMathisen:
- Altered length of subformat from 20 to 16 bytes. Old version would read too many bytes from validated file.
- Altered setting of compName, old code would give arrayIndexOutOfBounds exception for some Wave files.


david-russo (approved):

david-russo:
This also fixes the wrong WAVE format being detected under some circumstances.

PR #120, opened 2016-10-03
EXPERIMENTAL - Factor out PDF module string constants

carlwilson:
- created class for PDF MessageConstants; and
  - factoring out all String constants.


david-russo:
I like how _out_ you've factored them, but at the same time I feel they could be even _further_ out. Like resource-file out. If we had them in such a file, it might also be useful as a source for generating the online documentation... and other language translations... the methods for working with which could all be generalised under jhove-core for the benefit of all modules... but I'm getting _waaay_ ahead of this pull request...

Regarding which, we might want to make a note to change the message constant names for their equivalent IDs once they've been created, to avoid them possibly falling out of sync with their underlying messages.

Anyway, nice work, certainly an improvement :) I see no immediate problems with this, besides having to update the line numbers in the wiki again... Maybe leave this until after the hack day?

jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/MessageConstants.java
```diff
@@ -0,0 +1,160 @@
+/**
```
pmay, line 1: Missing "Invalid RoleMap" error from https://github.com/openpreserve/jhove/blob/release-1.14/jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/StructureTree.java#L249

carlwilson: Yvonne caught another, I'm on it ;)

jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/PdfModule.java
```diff
@@ -109,21 +110,274 @@
...
+	private static final String PROP_NAME_NO_FULL_PAGE = DICT_KEY_NO_FULL_PAGE;
+	private static final String PROP_NAME_PAGE_CLIP = DICT_KEY_PAGE_CLIP;
+	private static final String PROP_NAME_PRINT_AREA = DICT_KEY_PRINT_AREA;
+	private static final String PROP_NAME_VIEW_AREA = DICT_KEY_VIEW_AREA;
+	private static final String PROP_NAME_VIEW_CLIP = DICT_KEY_VIEW_CLIP;
+	private static final String PROP_NAME_DIRECTION = DICT_KEY_DIRECTION;
+
+	private static final String DICT_KEY_FONT_DESCRIPTOR = "FontDescriptor";
+	private static final String DICT_KEY_STARTXREF = "startxref";
+	private static final String DICT_KEY_BASE_FONT = base + font;
```
david-russo: This code would be easier to read as a literal "BaseFont", cutting out the time it takes to confirm the contents of the variables, and allowing the key to be searched. They're also variables that are unlikely to need changing in the future, so I see little benefit in having them as variables. I think the same can be said for most of the variables declared on lines 116 to 121, with the exception of `crop_box`.

jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/PdfModule.java
```diff
@@ -109,21 +110,274 @@
 public class PdfModule
     extends ModuleBase
 {
+	public static final String MIME_TYPE = "application/pdf";
+	public static final String EXT = ".pdf";
+	
+	private static final String font = "Font";
+	private static final String crop_box = "CropBox";
```
david-russo: Perhaps `PROP_VAL_CROP_BOX` to match the others?

jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/PdfModule.java
```diff
@@ -109,21 +110,274 @@
...
+	private static final String base = "Base";
+	private static final String name = "Name";
+	private static final String file = "File";
+	private static final String plural_s = "s";
+
+	private static final String PDF_VER1_HEADER_PREFIX = "PDF-1.";
+	private static final String PDF_SIG_HEADER = "%" + PDF_VER1_HEADER_PREFIX;
+	private static final String POSTSCRIPT_HEADER_PREFIX = "!PS-Adobe-";
+	private static final String ENCODING_PREFIX = "ENC=";
+    private static final String NO_HEADER = "No PDF header";
```
david-russo: Consistent space indentation to match the rest/majority of the file.

jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/Tokenizer.java
```diff
@@ -55,6 +55,9 @@
         '\u00f8','\u00f9','\u00fa','\u00fb','\u00fc','\u00fd','\u00fe','\u00ff'
     };
 
+    private static final String empty = "";
+    private static final String stream = "stream";
```
david-russo: `EMPTY`, `STREAM`?

PR #147, opened 2016-10-10
EXPERIMENTAL - Factor out JPEG module string constants

### LINKED ISSUE #139: Externalise JPEG module error messages

carlwilson:
- created class for JPEG MessageConstants; and
- factoring out all message String constants.

Closes #139


carlwilson:
This really requires some JPEG test files to ensure that the error output hasn't been altered.

carlwilson:
Hi @rosetta-development a quick review of this to ensure that I've not inadvertently changed any error messages. Once you've briefly double checked I'll merge this.

PR #150, opened 2016-10-11
EXPERIMENT XML & UTF-8 message constants

### LINKED ISSUE #144: Externalise XML module validation error messages

tledoux:
- Add UtfMessageConstants
- Add MessageContants for XML
- Extract strings
- Caveat : most messages from XML derives from SAXParseException ...

Fixes #144


tledoux:
The checks fail because of the "SaxParseException" prefix added.

carlwilson:
Hi @tledoux I've cloned this PR at #157 so that I can "push-button" update it. We need a few test cases before merging these PR's but we can work on that following the hack day. If that's OK I'll close this PR. Thanks very much and am about to create local clones of the other PRs.

carlwilson:
This seems to introduce an unexpected regression test in the Travis build when testing the output of running the XML-module over a copy of its own config file: https://travis-ci.org/openpreserve/jhove/jobs/166893457#L3016. The message is of a similar form but with the "SaxParseException:" embellishment. We can chat about this and how much it matters, this might be better.

tledoux:
Indeed, the MessageConstants class adds the "SaxParseException:" prefix in order to locate the use of the message. Otherwise, as discussed during the call, the ErrorMessage or the InfoMessage is directly the message generated by the exception of the SaxParser which seems very complex to manage since its implementation is a runtime information which can be overide by the edu.harvard.hul.ois.jhove.saxClass variable.

PR #151, opened 2016-10-11
PDF-Hul: Bug in skipIISBytes and PdfModule.getObject

pmay:
Many documents returning Invalid Page Dictionary Object seem to be the result of a bug in PdfFlateInputStream.skipIISBytes which is miscalculating the number of bytes to skip when the requested skip number is larger than the remaining buffer size.

In particular, this seems to relate to Page Trees encoded in stream objects where the root page starts beyond one buffer's worth of data.

The added InvalidPageDictionary.pdf file exemplifies this problem.

Note solving this problem results in "Improperly Constructed Page Tree" being returned by JHOVE. This seems to be being caused by JHOVE not correctly setting the object index for objects extracted from streams, meaning that when PageTreeNode.nextPageObject (line 197) tries to check if it's already visited a node, it fails (essentially it compares index -1 to index -1).


jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/PdfModule.java
```diff
@@ -2426,7 +2426,10 @@ protected PdfObject getObject (int objIndex, int recGuard)
                     }
                 }
                 /* And finally extract the object from the object stream. */
-                return ostrm.getObject (objIndex);
+                /* Need to ensure the object number is set */
+                PdfObject obj = ostrm.getObject (objIndex);
+                obj.setObjNumber (objIndex);
```
david-russo, line 2431: Setting the object number can be safely moved into the `ObjectStream.getObject()` method for the benefit of any other callers.

david-russo (approved):
Looks good to me.

carlwilson:
This is fixed by #188

PR #153, opened 2016-10-11
EXPERIMENT HTML message constants

### LINKED ISSUE #138: Externalise HTML Module validation error messages

tledoux:
Create class for HTML MessageConstants
Factoring out message String constants.
Beware some java sources are derived from jj files.
Fixes #138.


carlwilson:
Cloned in #158 for easy updating.

PR #159, opened 2016-10-11
Fix highlight travis regression errors

carlwilson:


carlwilson:
Merged other PR's unintentionally so replaced this with #160 and closed.

PR #185, opened 2017-01-10
Fix validation of WAVE files over 2 GB

### LINKED ISSUE #167: Support WAV's larger than 2GB

david-russo:
Chunk sizes were being incorrectly cast from `long` to `int`, producing inaccurate size calculations for chunks over the maximum value of `int` (2,147,483,647). Fixes #167.


carlwilson:
Fixed by #188

Issue #71, opened 2016-05-02
edu.harvard.hul.ois.jhove.ModuleBase: skipBytes() might not skip all requested bytes


RogerMathisen:
For large Wavefiles (>100MB) it happens that not all bytes in the DataChunk are skipped as expected in the method:

public long skipBytes(DataInputStream stream, long bytesToSkip, ModuleBase counted)

this seems to be because the call:

long n = stream.skip(bytesToSkip);

Actually might skip fewer bytes than requested (This is also stated in the Java Documentation). If this occurs it will most probably cause the parsing of the Wave file to fail, since the pointer to the next chunk will be placed inside the DATA chunk.

To avoid this problem the "long n = stream.skip(bytesToSkip);" call could be placed inside a loop that continues until all the desired bytes are skipped, or no more bytes kan be skipped (ie n=0).

smmorrissey:
Sheila +1

pmay:
+1

asciim0:
+1

friesey:
+1

PR #12, opened 2014-12-02
3 small fixes in Pdf-module

hak223sve:
1. Better handling Pdf-files where image heigth and width are PdfIndirectObjects
2. Better handling of Pdf-files with "empty" hex strings
3. Better handling of Pdf-files where form-fields are PdfIndirectObjects

PR #188, opened 2017-03-01
Merging PRs and fixing issues

### LINKED ISSUE #96: CrossRefStream incorrectly assumes /Index value is a 2 element array
The "isValid" method of CrossRefStream is hard coded to assume that an Index element, if present, is an array of exactly 2 integers.  According to the specification, the Index element is "an array containing a pair of integers for each subsection in this section." (http://wwwimages.adobe.com/content/dam/Adobe/en/devnet/pdf/pdfs/pdf_reference_archives/PDFReference15_v6.pdf, page 83 and PDF versions 1.6 and 1.7)  Documents that have more than one subsection fail validation as a result.  PdfModule.readXRefStreams appears to incorporate this assumption and does not validate object numbers against the actual object ranges specified in the index, but instead looks for numbers between 0 and the (meaningless?) value of CrossRefStream.getNumObjects()

BezrukovM:
closes #96, #71, #167
contains #97, #12, #185
also contains #151, but only code (without two test .pdf files)

Fixed Outlines parsing bug, fixed name tree parsing error.

### DIFF
--- a/jhove-core/src/main/java/edu/harvard/hul/ois/jhove/ModuleBase.java
+++ b/jhove-core/src/main/java/edu/harvard/hul/ois/jhove/ModuleBase.java
@@ -1366,11 +1366,20 @@ public abstract class ModuleBase
 			  ModuleBase counted) 
             throws IOException
     {
-		long n = stream.skip(bytesToSkip);
-        if (counted != null) {
-            counted._nByte += n;
+        long retVal = 0;
+        long bytesLeft = bytesToSkip;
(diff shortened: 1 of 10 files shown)


david-russo:
You might consider merging #169 instead of #97. It contains the refactoring and some additional improvements.

carlwilson:
To be fair, this is probably more my fault, I was working off the milestone list which was created before  #169 @BezrukovM  I can help with the git coercion here if that helps?

PR #117, opened 2016-09-09
-Modified skipBytes() Issue #71.

RogerMathisen:
Fix to handle skipping past the end of the file.

PR #194, opened 2017-03-17
Fix recently broken ModuleBase.skipBytes method

### LINKED ISSUE #71: edu.harvard.hul.ois.jhove.ModuleBase: skipBytes() might not skip all requested bytes

### LINKED ISSUE #193: JHOVE returns ErrorMessage: Unexpected end of file on WAV files
JHOVE 1.16 returns the error
ErrorMessage: Unexpected end of file
on all the wav files I tried so far both files over 2 GB and under 2 GB.

Here is a sample message:
ReportingModule: WAVE-hul, Rel. 1.4 (2017-03-14)
  LastModified: 2016-05-19 18:21:38 EDT
  Size: 49926728
  Format: WAVE
  Status: Not well-formed
  SignatureMatches:
   WAVE-hul
  ErrorMessage: Unexpected end of file
   Offset: 50057144
  MIMEtype: audio/x-wave

Can't attach any files as they are over the GitHub limit of 10 mb

JHOVE 1.14 works correctly on WAV files under 2GB but finds WAV files over 2 GB not well formed 
 ReportingModule: WAVE-hul, Rel. 1.3 (2007-12-14)
  LastModified: 2017-01-19 13:18:41 EST
  Size: 2351006108
  Format: WAVE
  Status: Not well-formed
  SignatureMatches:
   WAVE-hul
  ErrorMessage: Invalid character in Chunk ID: Character = 0x03
   Offset: 131075
  MIMEtype: audio/x-wave
  Profile: PCMWAVEFORMAT

david-russo:
Broken during incorrect merge of PR #117. Each loop of the skip method was skipping the same amount of bytes, not subtracting the amount already skipped. Reverted to original pull request code then refactored. Fixes #71, fixes #193.

Since this bug is in the core JHOVE code and can affect all modules, I'd say it warrants a minor version release.

### DIFF
--- a/jhove-core/src/main/java/edu/harvard/hul/ois/jhove/ModuleBase.java
+++ b/jhove-core/src/main/java/edu/harvard/hul/ois/jhove/ModuleBase.java
@@ -1295,7 +1295,7 @@ public abstract class ModuleBase
     }
 
     public static double readDouble (RandomAccessFile file, boolean endian)
-        throws IOException 
+        throws IOException
     {
         double f = 0.0F;
         if (endian) {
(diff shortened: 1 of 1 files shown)


carlwilson:
I'd agree regarding that minor release, I'll prepare this and merge as soon as possible.

PR #231, opened 2017-04-27
FIX - Typo

samalloing:


carlwilson:
Thank you @samalloing I've been pernickety and changed the title of the PR, I like to lay them out like so #225 the idea is that I can use the FIX - or DOC - markers to autogen the changelog in time.

Issue #237, opened 2017-04-27
Configuration logging settings not honoured


carlwilson:
Adding the <logLevel> element to the config files has no effect to the apps logging. This is caused by a logic bug where the config settings are only read when a `_logLevel` variable was null, and it never can be.

PR #238, opened 2017-04-27
FIX - Logging Config

carlwilson:
- fixed logging option handling so that logging config file changes work;
- tested param level setting also works; and
- made SecurityException handling a little more explicit.


carlwilson:
fixes #237

PR #263, opened 2017-07-19
Fix NPE preventing XRefStream objects being read

carlwilson:
A NullPointerException was being thrown when trying to validate the
Index entry of any CrossRefStream dictionary.

The NPEs were hidden by the method's general Exception handling, and even
though the containing method was returning false, and ending file
validation, no errors of any kind were being reported.

This resulted in any PDF with a CrossRefStream object being reported as
"Well-formed and valid" though most of the file's contents would remain
unchecked.

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -10,21 +10,20 @@ Versions 1.12 onwards of JHOVE released by the Open Preservation Foundation.
 
 JHOVE 1.16
 ----------
-2016-07-20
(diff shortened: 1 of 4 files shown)


carlwilson:
1.16 hotfix for #258 to fix a PDF module bug.

david-russo (approved):

carlwilson:
Includes and closes #258

PR #283, opened 2017-10-30
REFACT - TIFF Module error messages

carlwilson:
- renamed INF and ERR messages in line with "convention"; and
- replaced string occurences in code with MessageConstant values.

### DIFF
--- a/jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/TiffModule.java
+++ b/jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/TiffModule.java
@@ -435,7 +435,7 @@ public class TiffModule extends ModuleBase {
             byte ch0 = _raf.readByte();
             byte ch1 = _raf.readByte();
             if (ch0 != ch1 || (ch0 != 0X49 && ch0 != 0X4D)) {
-                throw new TiffException("No TIFF header: " + (char) ch0
+                throw new TiffException(MessageConstants.ERR_TIFF_HEADER_MISSING + (char) ch0
                         + (char) ch1, 0);
             }
             inHeader = false;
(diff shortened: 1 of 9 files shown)


jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/TiffModule.java
```diff
@@ -941,26 +934,24 @@ protected void checkValidity(TiffIFD ifd, RepInfo info)
...
+                reportInvalid(MessageConstants.ERR_PAL_COL_SPP_NE_1 + samplesPerPixel, info);
             }
             int len = (1 << bitsPerSample[0]);
             if (colormapBitCodeValue.length < len) {
-                reportInvalid("Insufficient ColorMap values for "
-                        + "palette-color: " + colormapBitCodeValue.length + "<"
-                        + len, info);
+                reportInvalid(MessageConstants.ERR_COL_MAP_MISS_VALS +
+                		      colormapBitCodeValue.length + "<"  + len,
+                		      info);
```
david-russo: Mixed tabs and spaces. Does Eclipse not have a "use existing indentation style"-type setting? If not, I can't recommend IntelliJ enough, for that and many other quality-of-life improvements. One of my own life's few regrets is trying to stick with Eclipse for as long as I did while IntelliJ was an option.

jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/tiff/GPSInfoIFD.java
```diff
@@ -15,6 +15,7 @@
 public class GPSInfoIFD
     extends IFD
 {
+	private static final String period = ".";
```
david-russo: Unnecessary constant. It's less code and clearer to read without this variable, and it doesn't appear to provide any maintenance benefits since the value isn't going to change. If it were to, the name of the variable would ideally be something more along the lines of `GPS_VER_SEPARATOR` so that the value could change without also needing the name to.

david-russo (commented):
Good stuff.

PR #284, opened 2017-10-31
REFACT - AIFF Module message constants

carlwilson:
- factored string constants to `MessageConstants` class for aiff and iff modules; and
- bumped AIFF module version.


david-russo (approved):
LGTM

PR #292, opened 2017-11-06
REFACT - External modules to discrete project

carlwilson:
- created new Maven module for external JHOVE modules: `jhove-ext-modules`;
- moved GZIP, PNG, and WARC modules to `jhove-ext-modules`;
- moved Maven dependencies for external modules to `jhove-ext-modules`;
- added `jhove-ext-modules` as Maven dependecy for `jhove-installer`;
- added new optional installer pack for jhove external modules;
- added alternative config for external modules at `jhove-installer/src/main/config/ext-modules/jhove.conf`; and
- classpath in executions scripts now `JHOVE_HOME/bin/*` rather than specific jar in order to pick up external modules jar.


jhove-installer/src/main/scripts/jhove
```diff
@@ -57,7 +57,7 @@ JHOVE_HOME="$( cd "$(dirname "${SCRIPT}" )" && pwd )"
 cd "${CWD}"
 
 # Create Java class path
-CP="${JHOVE_HOME}/bin/jhove-apps-${JHOVE_VERSION}.jar"
+CP="${JHOVE_HOME}/bin/*"
```
david-russo, line 60: These changes may impact people who upgrade JHOVE in-place, without it being obvious that something's gone wrong. Old versions of `jhove-apps-*.jar` will remain in the `bin` directory (as they currently do), but now with multiple JARs on the classpath, the application may just resolve the first `jhove-apps-*.jar` it comes across with the necessary classes in it. If it's ordered alphabetically, it will use the oldest version of the `jhove-apps-*.jar` in the `bin` directory instead of the latest, or previously installed, version.

carlwilson: OK, I ran into this issue and the next PR contains a fix for this but you might not like it ;) It [cleans out the bin directory on installation now](https://github.com/openpreserve/jhove/blob/rc/jhove-1.18-rc/jhove-installer/src/main/izpack/install.xml#L64)

carlwilson: A (possibly) more elegant solution is easy enough by dividing up the classpath a little and using a substitution in the installer instead.

david-russo: Ha, nah, that's not *too* bad. *I* was contemplating purging the whole installation folder except for the `conf` directory. No need to be that rash though. We should still probably provide a warning that `bin` will be purged in an upgrade. Or better yet, can we do a more targeted clean-up? Only files of this pattern: `jhove-apps-*.jar`?

david-russo: Divide it in what way?

david-russo:
You may also want to update the Project Structure documentation in the README.

tledoux (approved):
Elegnt way of splitting out.

The jwat.version property could be removed from the general pom.xml.
Except from the docuemntation to update, LGTM !!!

README.md
```diff
@@ -59,8 +60,8 @@ JHOVE is currently being maintained by the
 
 Pre-requisites
 --------------
- 1. Java JRE 1.6
```
david-russo: The two trailing spaces were a line-break.

carlwilson: Ahhh, I suspect my editor may have had it's way here will check.

PR #317, opened 2018-03-16
FIX - PDF minor version number checking.

### LINKED ISSUE #207: PDF version checking not correct
PDF version checking is only carries out against M=1, leading to invalid version entries such as 1.9 being well-formed and valid. Test case T01_002 from the ipres test corpus replaces the valid 1.6 version entry in the header with 1.9. The JHOVE outcome is well-formed and valid, even though version 1.9 does not exist.

carlwilson:
- added test for invalid minor version number;
- added MessageConstant for invalid minor version;
- bumped PDF Module release details for release candidate;
- refactored `PdfHeader` class;
- JavaDoc for `PdfHeader`;
- unit tests and test resources for `PdfHeader`;
- added JHOVE RC PDF Module details to baseline; and
- improved result logging for Travis tests.
Closes #207

### DIFF
--- a/jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/PdfModule.java
+++ b/jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/PdfModule.java
@@ -1063,20 +1063,20 @@ public class PdfModule
 
     protected boolean parseHeader(RepInfo info) throws IOException
     {
-    	PdfHeader header = PdfHeader.parseHeader(_parser);
-    	if (header == null) {
+        PdfHeader header = PdfHeader.parseHeader(_parser);
+        if (header == null) {
             info.setWellFormed(false);
             info.setMessage(new ErrorMessage(MessageConstants.ERR_PDF_HEADER_MISSING, 0L));
             return false;
         }
-		if (!header.isVersionValid()) {
-			info.setValid(false);
-			info.setMessage(new ErrorMessage(MessageConstants.ERR_PDF_MINOR_INVALID, 0L));
-		}
-		_version = header.getVersionString();
-		_pdfACompliant = header.isPdfACompliant();
+        if (!header.isVersionValid()) {
+            info.setValid(false);
+            info.setMessage(new ErrorMessage(MessageConstants.ERR_PDF_MINOR_INVALID, 0L));
+        }
+        _version = header.getVersionString();
+        _pdfACompliant = header.isPdfACompliant();
         info.setSigMatch(_name);
-		return true;
+        return true;
     }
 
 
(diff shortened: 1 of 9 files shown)


jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/PdfModule.java
```diff
@@ -1065,83 +1063,20 @@ protected final void initParse()
...
-            if (ctlcnt < 4) {
-                _pdfACompliant = false;
-            }
-        }
-        catch (Exception e) {
-            // Most likely a ClassCastException on a non-comment
-            _pdfACompliant = false;
-        }
-        return true;
+		if (!header.isVersionValid()) {
```
david-russo: Mixed indentation white-space.

jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/PdfHeader.java
```diff
@@ -0,0 +1,184 @@
+/**
+ * 
+ */
```
david-russo: Empty Javadoc.

jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/PdfHeader.java
```diff
@@ -0,0 +1,184 @@
...
+	 *         {@link PdfHeader#MAX_VALID_MAJOR_VERSION}. Otherwise false.
+	 */
+	public boolean isVersionValid() {
+		// Set minor version to one larger than maximum so invalid if parse
+		// fails
+		int minorVersion = MAX_VALID_MAJOR_VERSION + 1;
+		try {
+			minorVersion = getMinorVersion(this.versionString);
+		} catch (NumberFormatException nfe) {
+			// TODO : Do nothing for now, the version number is still invalid.
```
david-russo, line 66: TODO: Nothing?

jhove-modules/src/test/java/edu/harvard/hul/ois/jhove/module/pdf/PdfHeaderTest.java
```diff
@@ -0,0 +1,101 @@
+/**
+ * 
+ */
```
david-russo: Empty Javadoc.

jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/PdfHeader.java
carlwilson: The aim is to not break existing behaviour. This ensures that this case is handled as before as the minor version number is set too large and false is returned. I've left the TODO as I'm intending to polish this a little more, but it needs a new error messages and is outside of the PR scope.

asciim0: I wonder if in the long term view it is as easy as thinking about the minor version not being set too large, especially with PDF 2.0 now available. Couldn't it be instead checking the versionString against a list of currently available versions? So, as of today: 1.0, 1.1, 1.2, 1.3, ,1.4, 1.5., 1.6, 1.7, 2.0 ? 
Downside is that the list of available versions would have to be updated everytime a new version becomes available - but if the code instead states that the max correct minor version is 7, which is true for PDF 1.x, it would have to still be updated regarding the limitation of minor version for 2.x as there currently only is 2.0 (and this would again have to be updated once 2.1 becomes available).
Does that make any sense?

PR #318, opened 2018-03-16
FIX - Document catalog type checking

### LINKED ISSUE #210: Value for /Type in Document Catalog not validated
The document catalog can be found via an indirect reference in the trailer dictionary in key /Root. Per standard, the document catalog must have a /Type key of value /Catalog. T02-01_006 has a trailer pointing to obj 5 as the document catalog. While all other required entries for the Catalog dictionary are in place, the /Type value for the Catalog has been changed from the expected /Catalog to /Font. As a result the file cannot be rendered by Adobe Acrobat but JHOVE considers it well-formed and valid. There should be a consistency check for the indirect reference to the document catalog.

carlwilson:
- added test for document catalog `Type` key to `PdfModule`;
- added `MessageConstants` entries for catalog type checking errors;
- unit tests and test resources for functionality.
Closes #210

### DIFF
--- a/jhove-modules/src/test/java/edu/harvard/hul/ois/jhove/module/pdf/PdfModuleTest.java
+++ b/jhove-modules/src/test/java/edu/harvard/hul/ois/jhove/module/pdf/PdfModuleTest.java
@@ -1,0 +1,105 @@
+/**
+ * 
+ */
+package edu.harvard.hul.ois.jhove.module.pdf;
+
+import static org.junit.Assert.*;
+
+import java.io.File;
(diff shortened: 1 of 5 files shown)


jhove-modules/src/test/java/edu/harvard/hul/ois/jhove/module/pdf/PdfModuleTest.java
```diff
@@ -0,0 +1,105 @@
+/**
+ *
```
david-russo: Empty Javadoc.

Issue #173, opened 2017-01-02
ClassCastException: PdfDictionary cannot be cast to PdfStream


cchou:
## Dev Effort
5D

## Description
We encounter a class casting error while processing a PDF. The offending PDF can be downloaded at http://www.fcla.edu/daitss-test/files/03-2013.pdf.

I also downloaded the latest JHOVE version from openpreseve but still encounters the same error. 
The complete trackback is
./jhove -c conf/jhove.conf -m PDF-hul ~/Downloads/03-2013.pdf 
Jan 02, 2017 2:05:05 PM Jhove main
SEVERE: edu.harvard.hul.ois.jhove.module.pdf.PdfDictionary cannot be cast to edu.harvard.hul.ois.jhove.module.pdf.PdfStream
java.lang.ClassCastException: edu.harvard.hul.ois.jhove.module.pdf.PdfDictionary cannot be cast to edu.harvard.hul.ois.jhove.module.pdf.PdfStream
	at edu.harvard.hul.ois.jhove.module.PdfModule.parseTrailer(PdfModule.java:1031)
	at edu.harvard.hul.ois.jhove.module.PdfModule.parse(PdfModule.java:498)
	at edu.harvard.hul.ois.jhove.JhoveBase.processFile(JhoveBase.java:803)
	at edu.harvard.hul.ois.jhove.JhoveBase.process(JhoveBase.java:588)
	at edu.harvard.hul.ois.jhove.JhoveBase.dispatch(JhoveBase.java:455)
	at Jhove.main(Jhove.java:292)

PR #344, opened 2018-05-13
fix for #785 (https://github.com/openpreserve/jhove/issues/173)

cchou:
This fix the ClassCast Exception as described in issue https://github.com/openpreserve/jhove/issues/173. 

-Carol

### DIFF
--- a/jhove-apps/pom.xml
+++ b/jhove-apps/pom.xml
@@ -8,7 +8,7 @@
     <version>1.21.0-SNAPSHOT</version>
   </parent>
 
-  <artifactId>jhove-apps</artifactId>
+  <artifactId>jhove</artifactId>
   <packaging>jar</packaging>
 
   <name>JHOVE Applications</name>
--- a/jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/AProfile.java
+++ b/jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/AProfile.java
@@ -892,6 +892,8 @@ public final class AProfile extends PdfProfile
         }
         catch (Exception e) {
             return false;
+        }  catch (StackOverflowError e)  {
+             return false;
         }
         return true;
     }
--- a/jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/Name.java
+++ b/jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/Name.java
@@ -19,6 +19,12 @@ public class Name
     
     /** Returns true if it's within the PDF/A implementation limit */
     public boolean isPdfACompliant () {
-        return _value.getBytes().length <= 127; 
+        boolean compliant = false;
+        try {
+            compliant = _value.getBytes().length <= 127;
+        } catch (StackOverflowError e)  {
+            compliant = false;
+        }
+        return compliant; 
     }
 }
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
   </parent>
 
   <groupId>org.openpreservation.jhove</groupId>
-  <artifactId>jhove</artifactId>
+  <artifactId>jhove-app</artifactId>
   <version>1.21.0-SNAPSHOT</version>
   <packaging>pom</packaging>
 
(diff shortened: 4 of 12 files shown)


carlwilson:
Hi @cchou thanks for this. I'm looking at some of theses issues and JHOVE's error reporting. One issue is that we're now defining all messages in a single file: https://github.com/openpreserve/jhove/blob/integration/jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/MessageConstants.java rather than Strings spread across module files. There's also an issue with the change of artifact id in the top level POM, it's currently breaking the CI build. I can fix if you'd like, or you could make the changes? Thanks.

cchou:
Hi Carl,

Oh, I see, I did these fixes on 1.15 so I was merging them into 1.21. I catch all the rest of the message string but it appears I miss this one in merging PDFModule.java. Please go ahead and fix it on your side and I will pull them in.

In regards to POM.xml changes, it is needed for our project DAITSS. You can ignore the very last commit for POM.

Thanks,
-Carol

cchou:
HI @carlwilson ,

Per request, I have change the code to use MessageContant. I also restore pom setting back for the integration (will put in daitsss specific setting in a different branch). Please pull in the new changes.

Thanks,
-Carol

jhove-apps/pom.xml
```diff
@@ -8,7 +8,7 @@
     <version>1.21.0-SNAPSHOT</version>
   </parent>
 
-  <artifactId>jhove-apps</artifactId>
```
carlwilson, line 11: This is broken too now ;), it was fine as `jhove-aps`.

jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/AProfile.java
```diff
@@ -892,6 +892,8 @@ private boolean checkOutlineItem (PdfDictionary item)
         }
         catch (Exception e) {
             return false;
+        }  catch (StackOverflowError e)  {
```
carlwilson, line 895: This is interesting and suggests that the cause of the error lies in a recursive cycle when resolving one of the objects in `checkOutlineItem`, so some kind of check break might help here rather than just returning false.

pom.xml
```diff
@@ -9,7 +9,7 @@
   </parent>
 
   <groupId>org.openpreservation.jhove</groupId>
-  <artifactId>jhove</artifactId>
+  <artifactId>jhove-app</artifactId>
```
carlwilson, line 12: This one should be `jhove` please....

jhove-modules/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/Name.java
```diff
@@ -19,6 +19,12 @@ public Name ()
     
     /** Returns true if it's within the PDF/A implementation limit */
     public boolean isPdfACompliant () {
-        return _value.getBytes().length <= 127; 
+        boolean compliant = false;
+        try {
+            compliant = _value.getBytes().length <= 127;
+        } catch (StackOverflowError e)  {
```
carlwilson, line 25: This might be a red herring, I'm struggling to see how ` _value.getBytes().length <= 127` throws a stack overflow error as it's a Java `String` method. Will investigate.

carlwilson (commented):
Hi @cchou, I had a look at the error and the original file from #173. I've not quite had time to get my head around the stipulations of the spec vs. JHOVE's processing here. The main issue appears to be the assumption that the returned object would be a stream in the PdfModule, where you've added the type checks, so I've adopted that idea in this PR: #349. Ignore the new ASCII module files, they're other changes I'm working on that I inadvertently pushed.
It's your fix but I've:
 - dropped the `StackOverflowException` checks, I'm not sure they were needed in this context, but the one in `Parser` may be a good idea;
 - defined and used a [`MessageConstant.ERR_XREF_STRM_TYPE_INVALID`](https://github.com/openpreserve/jhove/pull/349/files#diff-8d88cc6be97f7e2559bbcad3b33583d6R171) for the error: 
 - simplified the type checks by throwing the exception first, meaning less indents and clearer code; and
 - dropped the POM changes.

If that's acceptable let me know and we can merge this and close.

cchou:
Carl, can you merge my new commit on Aug 21, commit 909f3571e564d9df03ac8ec011beddcd097ac0c9? This commit fix a JHOVE issue where it's dumping invalid MIX element for mix:exposureBiasValue. According to MIX spec, it's suppose to be a Rational type but JHOVE dump out double. A sample file to reproduce the error can be found at http://www.fcla.edu/daitss-test/files/zoca.jpeg. We also have our own ticket issue at https://github.com/daitss/core/issues/804.

Thanks,

carlwilson:
Hi @cchou, this PR is pretty stale now, and issue #173 has been closed for some time now. If there's something I'm missing let re-open it and let me know why in a comment.

PR #351, opened 2018-05-21
MAINT - Use real enums and parent class members

### LINKED ISSUE #373: MAINT - Use real enums and parent class members
## Dev Effort
2D

## Description
Java now supports native enumerations, JHOVE no longer needs to provide its own implementation. Some base classes aren't used as well as they can be, refactor some methods to a common base class.

carlwilson:
- moved repeatedly used `_ckSummer`, `_cstream`, and `_dstream` members to `ModuleBase`;
- tidied `ModuleBase` parameter handling;
- added a common method to set up module checksumming streams `ModuleBase.setupDataStream()`;
- `BytestreamModule` uses `ModuleBase.setupDataStream()`;
- replaced all types that extended `EnumerationType` with real Java `enum` types:
  - `AgentType`,
  - `AnalogDigitalFlagType`,
  - `ChecksumType`,
  - `DocumentType,`
  - `IdentifierType`,
  - `PropertyArity`,
  - `PropertyType`,
  - `SignatureType`,
  - `SignatureUseType`, and
- removed now defunct `EnumerationType`.

Closes #373


jhove-core/src/main/java/edu/harvard/hul/ois/jhove/IdentifierType.java
```diff
@@ -5,111 +5,113 @@
...
+
+	/******************************************************************
+	 * CLASS CONSTRUCTOR.
+	 ******************************************************************/
 
-    /******************************************************************
-     * CLASS CONSTRUCTOR.
-     ******************************************************************/
+	/**
+	 * Applications will never create SignatureTypes directly.
```
david-russo: Incorrect `*Type` referenced in comment. An old copy-paste error, though we probably don't need any of these comments anymore anyway since all `enum` constructors must be `private`. I'd do away with the large "CLASS CONSTRUCTOR"-style sign-boards as well, at least for simple enums.

jhove-core/src/main/java/edu/harvard/hul/ois/jhove/module/BytestreamModule.java
```diff
@@ -96,18 +87,7 @@ public final int parse (InputStream stream, RepInfo info, int parseIndex)
...
-	    ckSummer = new Checksummer ();
-            _cstream = new ChecksumInputStream (stream, ckSummer);
-            _dstream = getBufferedDataStream (_cstream, _je != null ?
-                        _je.getBufferSize () : 0);
-	}
-        else {
-            _dstream = getBufferedDataStream (stream, _je != null ?
-                                _je.getBufferSize () : 0);
-        }
+        setupDataStream(stream, info);
```
david-russo: The comment preceding this call explains code now hidden inside `setupDataStream`, so it would make more sense there instead.

jhove-core/src/main/java/edu/harvard/hul/ois/jhove/module/BytestreamModule.java
```diff
@@ -129,14 +109,14 @@ public final int parse (InputStream stream, RepInfo info, int parseIndex)
 	if (_nByte == 0) {
 	    info.setMessage (new InfoMessage (CoreMessageConstants.INF_FILE_EMPTY));
 	}
-        if (ckSummer != null) {
-	    info.setChecksum (new Checksum (ckSummer.getCRC32 (), 
+        if (_ckSummer != null) {
```
david-russo: I was going to suggest moving this checksum-related code to `ModuleBase` as well, but apparently there's already a `setChecksums` method there which you could replace this with. Might want to keep an eye out in other modules as well.

david-russo (commented):
Really good stuff :)

jhove-core/src/main/java/edu/harvard/hul/ois/jhove/module/BytestreamModule.java
carlwilson: I'm on the trail of a few Module similarities. The calculation of checksums is one, the proffering and checking of sigs is another. All of these were supposed to be abstract class tasks but cut and paste has prevailed.

PR #363, opened 2018-10-11
jhove goes up to 11, java11 🚀🎸🎶⑪⑾⒒⓫

willp-bl:
This compiles on java11...

...apart from when maven is packaging jhove-installer.  It is not able to get a dependency due to the client TLS implementation in java11 being strict (will apparently be corrected in January 2019 release of java11)

So the travis build for java11 is currently disabled :(

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ Gemfile.lock
 **/.settings
 **/.project
 **/.classpath
+
+.idea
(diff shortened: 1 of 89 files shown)


willp-bl:
The travis builds need to be restarted on travis, they failed due to network flakiness.  Unfortunately I can't do that

PR #381, opened 2019-01-19
Add javax.xml.bind as a dependency of modules

abdelq:
It is removed from Java since version 10 and moved to an external library instead. This allows to properly compile the project.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=917482


carlwilson:
Hi @abdelq thanks for this. I don't need to merge this as the dependencies were added as part of the last release. This Travis build highlights the duplicate dependency: https://travis-ci.org/openpreserve/jhove/jobs/598767782#L2685

PR #386, opened 2019-02-05
Adding support for SHA-256 message digest algorithm

escowles:
The [MessageDigest JavaDocs](https://docs.oracle.com/javase/9/docs/api/java/security/MessageDigest.html) say all implementations must support MD5, SHA-1, and SHA-256.  Since MD5 and SHA-1 are already supported, I added support for SHA-256.


carlwilson:
Hi @escowles and many thanks for this. Unfortunately it came just as we restructured the format modules while doing a little work to clean up the checksum calculation. Rather than try to sort out the git conflicts I've taken this work and created a new version of the PR: #497 that does the same thing. I'd be happy if you took a look at the PR to test it. I'm closing this PR on the back of the new one.

Issue #387, opened 2019-02-18
Incorret JHOVE Error „FocalPlaneResolutionUnit value out of range: 5“


rolandsuri:
## Dev Effort
5D

## Description
A value of five in the field "FocalPlaneResolutionUnit" is correct, as it follows the latest standard. According to the actual TIFF/EP-specification (ISO 12234-2:2001), the FocalPlaneResolutionUnit is a SHORT with values 1,2,3,4, or 5. In the documentation of Aware Systems, only the values 1,2, or 3 were allowed. The TIFF/EP-specifikation is newer and is referenced in the Module Info of TIF-hul. The TIFF/EP-specification also makes more sense: The 5 stands for Micrometer. As the optical imaging chips became smaller, the imager resolution can also me stated in terms of number of pixels per micrometer.

tledoux:
In fact, the problem is more complex because there is two specifications !!!
The TIFF/EP specifies the tag 37392 (named "FocalPlaneResolutionUnit") with 5 values
but the EXIF specifies the tag 41488 (also named "FocalPlaneResolutionUnit") with only 3 values !!!

This fact is very confusing and the files I have seen seem to indicate that many manufacturers don't respect the specifications or fill the tags erroneously... 
I try to come up with a list of values mostly compatible with the two spec. and consistent with the implementations...

MartinSpeller:
![](https://github.trello.services/images/mini-trello-icon.png) [Incorret JHOVE Error „FocalPlaneResolutionUnit value out of range: 5“ #387 - Assigned to TBA](https://trello.com/c/795JcEro/56-incorret-jhove-error-focalplaneresolutionunit-value-out-of-range-5-387-assigned-to-tba)

marhop:
Just FTR, something has changed here in the meanwhile: I have a TIFF file with a FocalPlaneResolutionUnit value of 4. JHOVE 1.20 complains, JHOVE 1.24 does not. (Also FTR: ExifTool recognizes value 4 in tag ExifIFD:FocalPlaneResolutionUnit (41488) as "mm".)

PR #394, opened 2019-02-28
Generate valid mix10 and mix20 metadata (various values are Rational).

tledoux:
The Brightness, ApertureBias and FlashEnergy properties are now Rational as needed for the versions 1.0 and 2.0 of the MIX schema.
Add unit test for all the new outputs.
Should also correct #387 : need for a sample file.


carlwilson (commented):
Nice, thanks for the tests :)

Issue #415, opened 2019-03-26
Position literals first in String comparisons


carlwilson:
There are 38 occurrences of this issue
Why is this an issue?
Since: PMD 3.3

Position literals first in comparisons, if the second argument is null then NullPointerExceptions can be avoided, they will just return false.

Example(s):
```java
class Foo {
  boolean bar(String x) {
    return x.equals("2"); // should be "2".equals(x)
  }
}
```
Codacy's list of all occurrences of the issue can be [found here](https://app.codacy.com/app/openpreserve/jhove/issues?bid=3402463&filters=W3siaWQiOiJMYW5ndWFnZSIsInZhbHVlcyI6W251bGxdfSx7ImlkIjoiQ2F0ZWdvcnkiLCJ2YWx1ZXMiOlsiRXJyb3JQcm9uZSJdfSx7ImlkIjoiTGV2ZWwiLCJ2YWx1ZXMiOltudWxsXX0seyJpZCI6IlBhdHRlcm4iLCJ2YWx1ZXMiOlsxMDYwXX0seyJpZCI6IkF1dGhvciIsInZhbHVlcyI6W251bGxdfV0=)

carlwilson:
Hi @nvanderperren I've chosen this as your first issue. It's a pet hate of mine. There's 38 instances of the problem, I'd be happier if you fix them a file at a time and then submit a pull request per file if that's OK? Here's an example in the [JHOVE window code](https://github.com/openpreserve/jhove/blob/integration/jhove-apps/src/main/java/edu/harvard/hul/ois/jhove/viewer/JhoveWindow.java#L710):
```java
if (_selectedModule.equals ("")) {
    return null;
}
```
It should read:
```java
if ("".equals(_selectedModule)) {
    return null;
}
```
This avoids the code throwing a null pointer exception when `_selectedModule` is null as the comparison method accounts for this.

Good luck

nvanderperren:
Thanks @carlwilson, will do!

nvanderperren:
I think I've solved all 38 occurences

carlwilson:
I concur, nice work.

nvanderperren:
yay! :tada:

carlwilson:
Fixed by #432, #433, #434, #435, #436, #437, #439, #440 and #441

PR #437, opened 2019-04-09
change position string literals in AIFF-Module / CommonChunk.java

nvanderperren:
See #415.

Not really related to this issue, but it's maybe more elegant to use a `switch` statement in stead of `if`'s?


carlwilson (commented):
Hi @nvanderperren might this be better as a switch case construct instead? https://www.geeksforgeeks.org/string-in-switch-case-in-java/

DOH, only just seen your earlier comment, sorry but feel free to change :)

carlwilson (commented):
Again, perhaps consider going switch case here: https://www.geeksforgeeks.org/string-in-switch-case-in-java/

Even worse here, was working on a single screen device with unfamiliar interface.

carlwilson (approved):
A thing of beauty, thanks very much.

Issue #424, opened 2019-03-26
Avoid instantiating Boolean objects; reference Boolean.TRUE or Boolean.FALSE or call Boolean.valueOf() instead.


carlwilson:
There are seven occurrences of this issue.
Why is this an issue?
Since: PMD 1.2

Avoid instantiating Boolean objects; you can reference Boolean.TRUE, Boolean.FALSE, or call Boolean.valueOf() instead.

Example(s):
```java
Boolean bar = new Boolean("true");		// unnecessary creation, just reference Boolean.TRUE;
Boolean buz = Boolean.valueOf(false);	// ...., just reference Boolean.FALSE;
```
The full list of occurrences can be found [here on Codacy](https://app.codacy.com/app/openpreserve/jhove/issues?bid=3402463&filters=W3siaWQiOiJMYW5ndWFnZSIsInZhbHVlcyI6W251bGxdfSx7ImlkIjoiQ2F0ZWdvcnkiLCJ2YWx1ZXMiOlsiRXJyb3JQcm9uZSJdfSx7ImlkIjoiTGV2ZWwiLCJ2YWx1ZXMiOltudWxsXX0seyJpZCI6IlBhdHRlcm4iLCJ2YWx1ZXMiOls5NjRdfSx7ImlkIjoiQXV0aG9yIiwidmFsdWVzIjpbbnVsbF19XQ==)

carlwilson:
Hi @nvanderperren this is just a quick one really and not my final choice for you. I needed a quick choice as I have to drive to an office now. I suspect it won’t take long and I’ll be ready with another issue in a couple of hours.

carlwilson:
Closed by #444

PR #444, opened 2019-04-10
FIX instantiating booleans

nvanderperren:
See #424 

I also removed some unnecessary boxing of variabels that were already instantiated as Long/Integer/Boolan/etc. and changed long if-else into switch.

There's also duplicate code in some of these files, for example in RepTreeRoot
```java
int[] ar = (int []) pVal;
                for (i = 0; i < ar.length; i++) {
                    node.add (new DefaultMutableTreeNode (ar[i]));
                }
```
or
```java
for (Object item : l) {
            if (null == ptyp) {
                // Simple objects just need a leaf.
                node.add (new DefaultMutableTreeNode (item, false));
            }
            else switch (ptyp) {
                case PROPERTY:
                    node.add (propToNode ((Property) item));
                    break;
                case NISOIMAGEMETADATA:
                    node.add (nisoToNode ((NisoImageMetadata) item));
                    break;
                default:
                    // Simple objects just need a leaf.
                    node.add (new DefaultMutableTreeNode (item, false));
                    break;
            }
        }
```

maybe a private (generic) helper method could reduce the duplicate code?

### DIFF
--- a/jhove-apps/src/main/java/edu/harvard/hul/ois/jhove/viewer/RepTreeRoot.java
+++ b/jhove-apps/src/main/java/edu/harvard/hul/ois/jhove/viewer/RepTreeRoot.java
@@ -407,91 +407,95 @@ public class RepTreeRoot extends DefaultMutableTreeNode
         int i;
         Object pVal = p.getValue ();
         PropertyType typ = p.getType ();
-        if (typ == PropertyType.INTEGER) {
-            int[] ar = (int []) pVal;
-            for (i = 0; i < ar.length; i++) {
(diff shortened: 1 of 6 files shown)


carlwilson (commented):
Hi @nvanderperren this is rather wonderful thanks. I agree about the generic helper method, do you fancy giving it a go?. This is going to require the full tests to verify so I'll get that into shape in the mean time.

nvanderperren:
Yeah, would like to write it. I'll probably start with it later this afternoon. Now in a meeting.

PR #460, opened 2019-07-03
Adds EPUB validation extension module based on W3C's EPUBCheck

karenhanson:
This change is to add the `EPUB-ptc` extension module version 1.0. The module wraps the [W3C EPUBCheck validation tool](https://github.com/w3c/epubcheck) as a JHOVE module. 

Some notes in advance of full documentation:
1. Tests use examples derived from IDPF's EPUB 3 [samples ](https://github.com/IDPF/epub3-samples)and [test suite](https://github.com/IDPF/epub-testsuite) on GitHub and National Library of the Netherlands / Research [epubPolicyTests](https://github.com/KBNLresearch/epubPolicyTests), also on GitHub. These may be useful for additional test material.
2. The `checkSignatures()` method is based on the Library of Congress [EPUB2](https://www.loc.gov/preservation/digital/formats/fdd/fdd000278.shtml#sign)and [EPUB3](https://www.loc.gov/preservation/digital/formats/fdd/fdd000308.shtml#sign) documentation. For this `valid` is always set to `UNDETERMINED`. This check only looks at "magic numbers" and file extension. If the signature matches, it sets "well formed" to true. Note that it does not perform the additional package checks that contribute to "well formed"-ness in `parse()` so there is a small inconsistency there.
3. Well-formed/valid status is currently dependent on the messages returned by EPUBCheck (except when a non EPUB is passed and it throws an exception see [this issue](https://github.com/w3c/epubcheck/issues/1050)). The full list of possible messages and their severity according to EPUBCheck can be [seen here](https://github.com/w3c/epubcheck/blob/master/src/test/resources/com/adobe/epubcheck/test/command_line/listSeverities_expected_results.txt).
    * an EPUB is well formed if it has no messages with `(severity="FATAL") or (severity="ERROR" and message-id begins with "PKG-")`. In other words, fatal errors and package errors make the file "Not well formed".
    * If well formed, an EPUB is not valid if it has any messages with `severity="ERROR"`.
4. The EPUBCheck provides a lot of interesting properties that I've included in the JHOVE report. Some of them may affect preservation decisions, so I thought they seemed useful. For example, the presence of encryption, external fonts, or resources that are stored remotely but visually or aurally embedded in the EPUB. The result is an unusually large property/value section! You can review what is included [here](https://github.com/karenhanson/jhove/blob/f5cd4da535043ccacd0ea337f8d1a63e3d3c3fec/jhove-ext-modules/src/main/java/org/ithaka/portico/jhove/module/EpubModule.java#L362-L390) and [here](https://github.com/karenhanson/jhove/blob/epub-ext/jhove-ext-modules/src/main/java/org/ithaka/portico/jhove/module/epub/JhoveRepInfoReport.java#L401-L413).

I look forward to your comments, and especially welcome any feedback on the rationales described above!


jhove-ext-modules/pom.xml
```diff
@@ -41,9 +42,15 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+            <argLine>-Xss1024k</argLine>
```
carlwilson, line 49: So this is the additional argument for 32bit JVMs. I'm a little curious as to whether this would affect Travis builds as it appears that both of the JVMs on Travis are 64bit. Am also curious if the failure is determinate, i.e. it fails consistently without this arg.

jhove-ext-modules/src/main/java/org/ithaka/portico/jhove/module/EpubModule.java
```diff
@@ -0,0 +1,567 @@
...
+                fontPropertiesList.add(generateProperty(PROPNAME_FONT, fontProps));
+            }
+        }
+        fontPropertiesList.remove(null);
+
+        return fontPropertiesList;
+
+    }
+
+    /**
```
carlwilson, line 439: These generate property methods appear to be useful, but they're also quite generic and might be useful to other module devs. While I don't propose moving them yet it looks like a useful maintenance / refactoring task for a later date.

carlwilson (commented):
There seems to be an issue with the ordering of elements within the XML report. I'm not sure how serious it is and whether the test framework needs fixing or the module. In essence the XML reports are identical but the elements are in a different order. My guess is this has it's roots in the "multi-threadedness" of the underlying EPUB library.

jhove-ext-modules/pom.xml
karenhanson: The behavior was quite strange. It seemed that one Travis build or the other would fail but never both or none - I'm not sure if that was just a coincidence. They seemed consistent in their failure or success when I repeated the same Travis build. Running them on a small test Linux instance I had they were much less consistent - if I ran them a few times, eventually they would build. I wonder if using a 32-bit JVM simply increases the chance of failure, but the failure is not impossible in the 64-bit environment for some reason.

karenhanson:
I dug in to the property ordering. I think it might be because `EpubModule.generateProperties()` is returning a `Set` not a `List`. It's not fundamentally wrong, but to make it easier to compare files, I will change to a `List` at the top level of the property/value tree and then we can discuss whether it should use `Lists` instead of `Set`s everywhere or not.

carlwilson (approved):

carlwilson:
FYI @karenhanson am aware of the conflict and how to solve it. Am just waiting on merging this as the conflicts can only be in the test scripts and as this branch / PR is a new module it's easier to untangle conflicts here. It'll get merged this week.

PR #465, opened 2019-07-17
Change method scope to support extending module classes

karenhanson:
Portico is working to upgrade to 1.23 - we currently use a modified older version of JHOVE. There are several method/property scope changes would that allow us to continue to use some local module extensions without having to maintain a customized version of JHOVE. If they are benign (or even useful) for other users, it would be great if we could merge these.  

Question for reviewer(s): one of the changes is in the `xml-hul` module. Should I increment the version? and if so to a SNAPSHOT version?

Thanks!

### DIFF
--- a/jhove-core/src/main/java/edu/harvard/hul/ois/jhove/handler/XmlHandler.java
+++ b/jhove-core/src/main/java/edu/harvard/hul/ois/jhove/handler/XmlHandler.java
@@ -897,10 +897,15 @@ public class XmlHandler
     }
 
 
-    /** Checks if a property would produce an empty XML element, and
-     *  returns true if it would.
+    /**
+     * Checks if a property would produce an empty XML element, and returns true if
+     * it would.
+     * 
+     * @param property
+     * @param arity
+     * @return
      */
-    private boolean isPropertyEmpty (Property property, PropertyArity arity) {
+    public boolean isPropertyEmpty (Property property, PropertyArity arity) {
         try {
             if (arity.equals (PropertyArity.SET)) {
                 Set propSet = (Set) property.getValue ();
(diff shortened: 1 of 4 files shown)


jhove-core/src/main/java/edu/harvard/hul/ois/jhove/handler/XmlHandler.java
```diff
@@ -897,10 +897,15 @@ else if (arity.equals (PropertyArity.ARRAY)) {
...
+    /**
+     * Checks if a property would produce an empty XML element, and returns true if
+     * it would.
+     * 
+     * @param property
+     * @param arity
+     * @return
      */
-    private boolean isPropertyEmpty (Property property, PropertyArity arity) {
+    public boolean isPropertyEmpty (Property property, PropertyArity arity) {
```
tledoux, line 908: Shouldn't this method be static also ?

jhove-core/src/main/java/edu/harvard/hul/ois/jhove/handler/XmlHandler.java
```diff
@@ -4406,10 +4411,14 @@ private void writeAESTimeRange (String baseIndent,
...
+    /**
+     * Clean up a URI string by escaping forbidden characters. 
      * We assume (perhaps dangerously) that a % is the start of
-     * an already escaped hexadecimal sequence. */
-    private String cleanURIString (String uri) {
+     * an already escaped hexadecimal sequence.
+     * @param uri
+     * @return
+     */
+    public String cleanURIString (String uri) {
```
tledoux, line 4421: Shouldn't this method be static ?

tledoux (commented):
My feeling is that the 4 methods (encodeContent, encodeValue, isPropertyEmpty, cleanURIString) are really utility functions. It would be easier for maintenance to put them all in a Utility class, so we know others may use them and we don't mess with them.

jhove-core/src/main/java/edu/harvard/hul/ois/jhove/handler/XmlHandler.java
carlwilson: I'd not thought that clearly about it before @tledoux but I think you're right. Will pr a refactor.

carlwilson:
I'd not thought that clearly about it before @tledoux but I think you're right. Will pr a refactor.

carlwilson:
Included in #538

PR #466, opened 2019-08-05
Add documentation for EPUB module

karenhanson:
This is a first draft of the documentation to go with the EPUB module. I added it to the third party section. Preview can be seen here (links will require manual addition of /jhove/ to the path until published):
https://karenhanson.github.io/jhove/modules/third-party/epub/
Please let me know if any other information is required.


carlwilson (commented):
This is a really good start, I think that it should cover the issue around the possible stack overflow error when using a 32bit JVM, though there won't be too many affected users.

karenhanson:
I've added a troubleshooting section, per your feedback. I have proposed a work around - please let me know if there is an easier way!

tledoux (approved):

carlwilson (approved):

PR #467, opened 2019-08-13
German translation of PDF error messages

SteffenR2D2:


carlwilson (commented):
Hi @SteffenR2D2 this is fine, I'll be able to review it properly a little later and suggest changes. I'm not approving or rejecting right now.

carlwilson (changes requested):
Hi @SteffenR2D2 , the file structure is fine, it's only the location and name that need changing. The file should live in the appropriate resources area for the particular module. In the case of the PDF module that's jhove/jhove-modules/pdf-hul/src/main/resources/edu/harvard/hul/ois/jhove/module/pdf/ErrorMessages_de.properties. I can do the renames if necessary.

PR #468, opened 2019-08-19
WAVE / IFF: Accurate file offsets and new validation checks

david-russo:
Core:
- In certain circumstances submessages were being dropped when creating JhoveMessages.

IFF (AIFF & WAVE):
- Check that chunk IDs only consist of characters in the printable ASCII range.
- Check that spaces do not precede printable characters in chunk IDs.
- Clarified error messages and improved offset reporting accuracy.

WAVE:
- Added reporting for unrecognized data in the top-level RIFF structure.
- Made the Table Length field of `ds64` chunks optional to better align with the specification.
- Reinstated WAVE-HUL-4 reporting which had been lost during refactoring.
- Corrected WAVE-HUL-15 from an Error to an Informational message.
- Retired WAVE-HUL-16, an unused duplicate of WAVE-HUL-19.
- Clarified error messages and greatly improved offset reporting accuracy.
- Moved all remaining message text into translatable resource files.
- Added test files to demonstrate new validation checks.


carlwilson (commented):
Nice work :) Changes were sensible and easy to follow the logic.

Will add a PR to fix the tests but that will be a brute force copy of the WAV file results as the PR changes the WAV output in almost all cases. Need to bump the module version also.

david-russo:
Thank you. Yeah, the version numbers should be bumped for WAVE _and_ AIFF.

PR #487, opened 2019-10-11
TIFF: Add ASCII validation for ASCII fields

david-russo:
- Added ASCII validation for ASCII fields
- Fixed problem reporting `InkNames` field values


jhove-modules/tiff-hul/src/main/java/edu/harvard/hul/ois/jhove/module/tiff/IFD.java
```diff
@@ -560,78 +560,69 @@ protected Property propertyHeader(String type, List entries)
...
-           Object[], not a String[] ... sigh. */
-        String [] strs = new String[nstrs];
-        ListIterator<String> iter = list.listIterator();
-        for (int i=0; i<nstrs; i++) {
-            strs[i] =  iter.next();
+
+        if (invalidAscii) {
+            _info.setMessage(new ErrorMessage(
+                    MessageConstants.TIFF_HUL_72, value));
+            _info.setValid(false);
```
tledoux, line 622: Hi I have mixed feelings with declaring invalid TIFF files just because of non ASCII (but ISO-8859-1) characters.
Indeed, I know, because this control doesn't exist, that we have thousand of images that have "Bibliothèque nationale de  France" in comment but are otherwise perfectly fine.

It would be nice if we could make this configurable like the `byteoffset=true` parameter to allow it to simply be a warning.
A parameter like invalidNonAsciiCharacters with the default value true will be great.

carlwilson: @tledoux I've been working on a more generic way of ignoring / downgrading / upgrading errors by id using a configuration file passed at run-time. It might prove a useful addition as it makes reporting more flexible. Will do my best to get the PR out for the first part of next week for review.

david-russo: Ha, yeah, we ourselves have _quite_ a few (somewhat questionable) "©" symbols peppered about our TIFF collections. That was actually the motivator for the addition of this check. If JHOVE had had it earlier we would have picked up on this issue sooner and been able to remedy it with a better alternative.

The TIFF 5 and 6 specs are very clear that these fields should only contain "7-bit ASCII". Rendering software has no way to know what character encoding is being used in these fields if not ASCII, and so no characters outside of the ASCII range can be relied upon to render predictably outside of the software which wrote them. Given that, I think marking such TIFFs as invalid is actually fairly desirable, from a preservation perspective, though I certainly sympathize about now needing to deal with the problems our past mistakes have made for us.

The optional parameter isn't a bad idea, if you'd like to implement it, but it does make me think how much better a more general solution would be, allowing us to selectively suppress _any_ messages using our cool newfangled message IDs! Alas... we currently set our validity statuses separately from our messages, so until message objects carry their own status modifiers, parameters may be the way to go.

Then again, a part of me also can't help but feel that by providing an explicit option for this check we may be implying that it's safe or reasonable to ignore this incompatibility under certain circumstances, which I don't believe is necessarily _true_. So I'm torn... but feel free to add what options you like :)

david-russo: @carlwilson Just now saw your reply. _Of course_ you're already working on it :)

PR #489, opened 2019-10-15
New WARC documentation

### LINKED ISSUE #307: Missing documentation for WARC module
## Dev Effort
2D

## Description
Hello,

JHOVE (at least since version 1.16.1) ships with a module called "WARC-kb" that validates, well, WARC files. It's really nice and even extracts a lot of information from a WARC file.

However, it seems there exists no documentation whatsoever for this module, neither on <http://jhove.openpreservation.org/modules/> nor on GitHub. Am I missing something?

Thanks,
Martin

samalloing:
This commit resolves #307

### DIFF
--- a/modules/gzip/index.html
+++ b/modules/gzip/index.html
@@ -1,0 +1,108 @@
+---
+title: GZIP-kb Module
+---
+<!DOCTYPE html>
+<html lang="en">
+{% include header.html %}
+<body role="document">
+
+{% include navbar.html nav=site.data.navbar %}
+<div class="container" role="main">
+<h1>HTML-hul Module</h1>
+<a class="name" name="introduction">
+<h2>1 Introduction</h2>
+</a>
+<p>
+  The GZIP-kb module recognizes and validates the Gzip (GNU zip) format.
+  [<a href="/references#gzip">GZip</a>].
+</p>
+<p>
+  The module is invoked by the:
+</p>
+<blockquote>
+  <pre>
+    jhove ... -m GZIP-kb ...
+  </pre>
+</blockquote>
+<p>
+  command line option.
+</p>
+<p>
+  The GZIP-kb module recognizes and validates <a href="/references#gzip">GZip version 4.3</a>. It also supports multiple member GZip files. This module uses the <a href="/references#jwat">JWAT</a> library for GZip parsing.
+</p>
+
+<p>
+This module doesn't have configurable parameters.
+</p>
+
+<a class="name" name="coverage">
+  <h2>2 Coverage</h2>
+</a>
+<p>
+  The GZIP-kb module recognizes and validates the following public profiles:
+</p>
+<ul>
+  <li>
+    <a href="/references#gzip">RFC 1952</a>
+  </li>
+</ul>
+
+<a class="name" name="well-formedness">
+  <h2>3 Well-Formedness</h2>
+</a>
+<p>
+  The GZip module doesn't check the well-formedness.
+</p>
+
+<a class="name" name="validity">
+<h2>4 Validity</h2></a>
+<p>
+ The following criteria must be met by a GZip file for JHOVE to consider it valid:
+</p>
+<ul>
+<li> The file is well-formed
+</ul>
+
+<a class="name" name="repinfo">
+<h2>5 Representation Information</h2></a>
+<p>
+  The MIME type is reported as: application/gzip [<a href="/references#rfc6713">RFC 6713</a>]. Application/x-gzip is also supported
+  
+</p>
+<p>
+  In addition to the standard JHOVE
+  <a href="/documentation#repinfo">representation information</a>, the following
+  GZip-specific properties are reported:
+</p>
+<ul>
+<li>
+    Property "GzipEntryProperties"
+    <ul>
+		<li>Property "Is non compliant" of type STRING</li>
+		<li>Property "Offset value" of type STRING</li>
+		<li>Property "GZip entry name" of type STRING</li>
+		<li>Property "GZip entry comment" of type STRING</li>
+		<li>Property "GZip entry date" of type STRING</li>
+		<li>Property "GZip entry compression method" of type STRING</li>
+		<li>Property "GZip entry operating system" of type STRING</li>
+		<li>Property "GZip entry header crc16" of type STRING</li>
+		<li>Property "GZip entry crc32" of type STRING</li>
+		<li>Property "GZip entry extracted size (ISIZE)" of type STRING</li>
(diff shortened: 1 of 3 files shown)


samalloing:
Hi @carlwilson 

The commit with the GZip documentation is automatically added to this pull request. I don't know how to separate them. Is this fine by you?

Now I'm going to work on the dutch translation

Thanks

Sam

modules/gzip/index.html
```diff
@@ -0,0 +1,108 @@
...
+title: GZIP-kb Module
+---
+<!DOCTYPE html>
+<html lang="en">
+{% include header.html %}
+<body role="document">
+
+{% include navbar.html nav=site.data.navbar %}
+<div class="container" role="main">
+<h1>HTML-hul Module</h1>
```
tledoux:
```suggestion
<h1>GZIP-kb Module</h1>
```

modules/gzip/index.html
```diff
@@ -0,0 +1,108 @@
...
+  The GZip module doesn't check the well-formedness.
+</p>
+
+<a class="name" name="validity">
+<h2>4 Validity</h2></a>
+<p>
+ The following criteria must be met by a GZip file for JHOVE to consider it valid:
+</p>
+<ul>
+<li> The file is well-formed
```
tledoux: You just stated before that the GZip module doesn't check the well-formedness. Do you mean the file is compliant with the RFC as suggested by the above property ?

modules/warc/index.html
```diff
@@ -0,0 +1,113 @@
...
+
+{% include navbar.html nav=site.data.navbar %}
+<div class="container" role="main">
+<h1>WARC-kb Module</h1>
+<a class="name" name="introduction">
+<h2>1 Introduction</h2>
+</a>
+<p>
+  The WARC-kb module recognizes and validates the WARC (Web ARChive) format.
+  [<a href="/references#warciso">WARC</a>]. It only validates the WARC file format and WARC headers, not the actual payload of the WARC records.
```
tledoux, line 17:
```suggestion
  [<a href="/references#warc">WARC</a>]. It only validates the WARC file format and WARC headers, not the actual payload of the WARC records.
```
Use here the link to the warc specification which can be directly used (as opposed to the ISO standard)

modules/warc/index.html
```diff
@@ -0,0 +1,113 @@
...
+  This module uses the <a href="/references#jwat">JWAT</a> library for WARC parsing.
```
tledoux, line 18:
```suggestion
  This module uses the <a href="/references#jwat">JWAT</a> library for WARC parsing.
It also expects the GZIP-kb module to be present in order to parse compressed WARCs (.warc.gz) which interweave WARC records and GZIP entries.
```

modules/warc/index.html
samalloing: Hi Thomas,

Thanks for the feedback! I'm was doubting about this myself. But my reasoning eventually was, that the general links (like this example) point to the ISO standard and specific links (for example the file extension information) to the publicly available specs.

I also point to the 2009 version (ISO28500:2009) while there is a newer version of 2017, but the module documentation points to the 2009 version so that's why I reference it.

So @tledoux do you want me to change this? @carlwilson do you have an opinion about this (or somebody else).

modules/warc/index.html
samalloing: It was difficult to verify this because jHove bundles these modules in one jar, but I don't think the GZIP-kb module is required. Because it uses the JWAT library for the GZip parsing. Can you explain why you want this change? @tledoux

modules/gzip/index.html
samalloing: Well this is something I have doubts about. But it is implied in the WARC and the same way it is implemented in GZip.

The compliancy is evaluatued in JWAT and it is used for validation and well-formedness.

So in the code it is stated:
* Well-formedness is empty and validation is not (this is WARC, but the same happens in GZip: https://github.com/openpreserve/jhove/blob/dd25d0ff559931a68604f35fcdae337ddd4aae19/jhove-ext-modules/src/main/java/edu/harvard/hul/ois/jhove/module/WarcModule.java#L73-L74
* Same status is used for well-formedness and validity: https://github.com/openpreserve/jhove/blob/dd25d0ff559931a68604f35fcdae337ddd4aae19/jhove-ext-modules/src/main/java/edu/harvard/hul/ois/jhove/module/WarcModule.java#L230-L231
* but the compliancy is checked in JWAT: https://github.com/netarchivesuite/jwat/blob/eb87798130394893c1aed809cbe2365a9a52737e/jwat-warc/src/main/java/org/jwat/warc/WarcRecord.java#L458. You'll find GZip: https://github.com/netarchivesuite/jwat/blob/eb87798130394893c1aed809cbe2365a9a52737e/jwat-gzip/src/main/java/org/jwat/gzip/GzipReader.java#L185

I don't know if this is clear and if you want me to change anything. But thanks for the feedback! @tledoux 

Sam

tledoux: IMHO, wellFormedness means the file respects the grammar of the format while validity is granted when the file follows the constraint of a standard or a profile.
In the case of Gzip there is no place for distinction between these two notions : so I suppose it should be stated that Well-Formedness (which is what is checked here) is equivalent to Validity for this module.

modules/warc/index.html
tledoux: It seems to me that compressed WARC files (.warc.gz) are very common in the field. So something has to be told about the handling of this "format". Indeed, .warc.gz are very different from, said, .tar.gz. The latter is just the compression of a tarball, the first is a sequence of gzip entries that when they are individually uncompressed build a series of warc records that made a warc file: the two formats are really interweaved.

modules/gzip/index.html
samalloing: OK Thanks Thomas. I'll change it  I'll write that well-formedness is checked and in Validity the Well-formdness and Validity are equivalent. @tledoux

modules/warc/index.html
samalloing: I looked more in the JWAT library and there the GZip file is checked and the compliancy as well. So I'll add a sentence about that in .warc.gz the GZip compliancy is checked as well with the JWAT library. 

Thanks for your feedback, Thomas! @tledoux

samalloing:
Hi @carlwilson 

I updated this pull request with the comments of @tledoux. 

Sam

carlwilson:
Hi @samalloing I had been following the conversation at least but was happy for yourself and @tledoux to discuss. Have begun the great merge up but will be at this into next week.

carlwilson (approved):

PR #499, opened 2019-10-18
FIX - Message constant niggles

carlwilson:
- added message properties file for core constants;
- corrected missed concatenation from `PDF-HUL-3`;
- added exception message to creation of `PDF-HUL-3` in `Destination`; and
- removed Java punctuation from `ASCII-HUL-1-SUB`.


tledoux (approved):
LGTM

PR #500, opened 2019-10-18
typo in CoremessageConstants

samalloing:
During the translation I noticed this weird sentence. When I look at the code I think what is meant is *for* and not *or*. So if you could confirm this, then I can change the translation

Thanks

Sam


carlwilson (approved):

carlwilson:
Confirmed @samalloing

Issue #418, opened 2019-03-26
Avoid unnecessary if..then..else statements when returning booleans


carlwilson:
There are 59 occurrences of this issue.
Why is this an issue?
Since: PMD 0.9

Avoid unnecessary if-then-else statements when returning a boolean. The result of the conditional test can be returned instead.

Example(s):

public boolean isBarEqualTo(int x) {

	if (bar == x) {		 // this bit of code...
		return true;
	} else {
		return false;
    }
}

public boolean isBarEqualTo(int x) {

   	return bar == x;	// can be replaced with this
}
Codacy's list of all occurrences can be [found here](https://app.codacy.com/app/openpreserve/jhove/issues?bid=3402463&filters=W3siaWQiOiJMYW5ndWFnZSIsInZhbHVlcyI6W251bGxdfSx7ImlkIjoiQ2F0ZWdvcnkiLCJ2YWx1ZXMiOlsiRXJyb3JQcm9uZSJdfSx7ImlkIjoiTGV2ZWwiLCJ2YWx1ZXMiOltudWxsXX0seyJpZCI6IlBhdHRlcm4iLCJ2YWx1ZXMiOls4OTRdfSx7ImlkIjoiQXV0aG9yIiwidmFsdWVzIjpbbnVsbF19XQ==)

nvanderperren:
On this now!

carlwilson:
Thanks @nvanderperren

PR #505, opened 2019-10-21
Fixed: duplicate if statements

nvanderperren:
Tries to solve #418


jhove-modules/html-hul/src/main/java/edu/harvard/hul/ois/jhove/module/html/HtmlDocDesc.java
```diff
@@ -344,11 +344,9 @@ private boolean checkElementContext(JHOpenTag elem, RepInfo info) {
 		 */
 		String name = elem.getName();
 		if (elementStack.isEmpty()) {
-			if ("html".equals(name)) {
-				return true;
-			}
-			// This shouldn't happen
-			return false;
+                    // This shouldn't happen
```
carlwilson: remove misleading comment

carlwilson (changes requested):
1 very minor change requested but that apart a thing of beauty, thanks. Went a bit operator blind checking the logic. Thanks muchly @nvanderperren

jhove-modules/html-hul/src/main/java/edu/harvard/hul/ois/jhove/module/html/HtmlDocDesc.java
nvanderperren: should be fixed!

carlwilson (approved):

Issue #422, opened 2019-03-26
Switch statements should have a default label


carlwilson:
There are 12 occurrences of this issue.
Why is this an issue?
Since: PMD 1.0

All switch statements should include a default option to catch any unspecified values.

Example(s):
```java
public void bar() {
    int x = 2;
    switch (x) {
      case 1: int j = 6;
      case 2: int j = 8;
      				// missing default: here
    }
}
```
A full list of occurrences on codacy can be [found here](https://app.codacy.com/app/openpreserve/jhove/issues?bid=3402463&filters=W3siaWQiOiJMYW5ndWFnZSIsInZhbHVlcyI6W251bGxdfSx7ImlkIjoiQ2F0ZWdvcnkiLCJ2YWx1ZXMiOlsiRXJyb3JQcm9uZSJdfSx7ImlkIjoiTGV2ZWwiLCJ2YWx1ZXMiOltudWxsXX0seyJpZCI6IlBhdHRlcm4iLCJ2YWx1ZXMiOls4OTFdfSx7ImlkIjoiQXV0aG9yIiwidmFsdWVzIjpbbnVsbF19XQ==)

nvanderperren:
I will clean this one now! 🚿

nvanderperren:
@carlwilson some errors in TextHandler.java of Jhove Core module: `reference to Module is ambiguous. both interface edu.harvard.hul.ois.jhove.Module in edu.harvard.hul.ois.jhove and class java.lang.Module in java.lang match`. Also errors of overridden methods that are not overriden.

Do I make a seperate issue of this? not sure if this is an important error or class...

tledoux:
Hi @nvanderperren, the errors about the ambiguous `Module` class come from the fact that you use a Java 9.0 or greater to compile Jhove. Indeed this version of Java introduced the notion of modules at the core of the language which hits the use of Module by Jhove. The easy fix is just to desambiguate using the full name each time it is refered to.

We definitively need to fix this in order to remain compatible. A separated issue will probably be better to handle it.

nvanderperren:
Ok, I create a separate issue.

Op di 22 okt. 2019 09:28 schreef tledoux <notifications@github.com>:

carlwilson:
@nvanderperren many thanks for this, the switch default clean up was on my list for this week. I can see where the issue has arisen regarding use of module but that is another issue. Thanks again.

PR #507, opened 2019-10-22
add default switch statements

nvanderperren:
tries to solve #422 
since it wasn't much, I also added `@param`, `@return`, `@throws` etc. to the Javadocs.


jhove-modules/html-hul/src/main/java/edu/harvard/hul/ois/jhove/module/HtmlModule.java
```diff
@@ -446,11 +450,7 @@ public int parse(InputStream stream, RepInfo info, int parseIndex) {
 			HtmlDocDesc docDesc = null;
 			switch (type) {
 			case HTML_3_2:
-			default:
```
carlwilson, line 449: @nvanderperren I'm missing something with this deletion? A quick explanation for the hard of understanding please.

carlwilson (commented):
I mostly wholeheartedly approve of these changes, there's one or two I halfheartedly approve of but recent PRs from others lead me to suspect I'm in a minority. Checking the appropriate style guidelines didn't strengthen my points ;).

Just the one comment inline, I suspect the problem is my understanding.

Finally I'll hold off this PR until a few earlier ones are merged. Many of the changes here are logical and I'd rather deal with conflicts in this branch / PR than others.

jhove-modules/html-hul/src/main/java/edu/harvard/hul/ois/jhove/module/HtmlModule.java
nvanderperren: the null values are not necessary because it's not used nor checked. Only the values that are assigned in the if or switch statements are used. But I was maybe a bit too eager.

nvanderperren:
Thanks @carlwilson . Next time, I'll create a new branch for the lacking Javadoc documentation and  a seperate PR for that issue.

PR #555, opened 2019-12-14
Issue554

tledoux:
Distinction between invariant and mult descriptions.

Correct tests to handle differences in locale.
Correct error message in JPEG.


carlwilson (approved):
Thanks @tledoux I'll create a new RC shortly.

Issue #559, opened 2020-01-16
JHOVE GUI fails to display TIFFMetadata properties


david-russo:
Where previously the JHOVE GUI would render the TIFFMetadata properties like so:
```
+ TIFFMetadata
  | ByteOrder: little-endian
  + IFDs
    | Number: 1
    + IFD
      | ...
```

JHOVE 1.24-RC4 now only displays:
```
+ TIFFMetadata
  | edu.harvard.hul.ois.jhove.Property@1a8922a
  | edu.harvard.hul.ois.jhove.Property@1409d1f
```

The command-line interface doesn't appear to have any trouble displaying the same properties.

Example file: https://github.com/openpreserve/jhove/blob/rel/jhove-1.22/test-root/corpora/examples/modules/TIFF-hul/AA_Banner.tif

Windows 7
JRE 1.8 x86

david-russo:
The issue also appears with certain properties in the output of other modules, such as WAVE-hul:
```
+ WAVEMetadata
  + AESAudioMetadata
  | ...
  + Samples
    + Sample
      | edu.harvard.hul.ois.jhove.Property@1aa8e0d
      | edu.harvard.hul.ois.jhove.Property@aa0f9d
      | ... 
```
https://github.com/openpreserve/jhove/blob/rel/jhove-1.22/test-root/corpora/examples/modules/WAVE-hul/8-Bit-Noise-1.wav

PR #562, opened 2020-02-07
FIX - JHOVE GUI property display issues

### LINKED ISSUE #558: JHOVE GUI fails to display info. for certain JPEG files
JHOVE 1.24-RC4 fails to display a RepInfo window in the GUI after choosing the JPEG-hul module and attempting to open certain JPEG files. The same files don't exhibit the issue in JHOVE 1.22.1. The command-line interface for 1.24-RC4 also doesn't seem to have trouble parsing the files.

Example file: https://github.com/openpreserve/jhove/blob/rel/jhove-1.22/test-root/corpora/examples/modules/JPEG-hul/20150213_140637.jpg

Windows 7
JRE 1.8 x86

carlwilson:
- added fix to `RepTreeRoot.addToNode()` that ensures `Property` elements are rendered correctly; and
- hack patch for `addArrayMembers()` to ensure that primitive `int[]`s and `long[]`s are converted to `Integer[]`s and `Long[]`s respectively.

Fixes #558 and #559

### DIFF
--- a/jhove-apps/src/main/java/edu/harvard/hul/ois/jhove/viewer/RepTreeRoot.java
+++ b/jhove-apps/src/main/java/edu/harvard/hul/ois/jhove/viewer/RepTreeRoot.java
@@ -6,6 +6,7 @@
 package edu.harvard.hul.ois.jhove.viewer;
 
 import java.text.DateFormat;
+import java.util.Arrays;
 import java.util.Date;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


jhove-apps/src/main/java/edu/harvard/hul/ois/jhove/viewer/RepTreeRoot.java
```diff
@@ -438,11 +439,25 @@ private void addArrayMembers(DefaultMutableTreeNode node, Property p) {
...
 			case LONG: {
-				addToNode(node, (Long[]) pVal);
+				if (pVal instanceof long[]) {
+					Long[] vals = Arrays.stream((long[])pVal) // IntStream
+							.boxed()				// Stream<Integer>
+							.toArray(Long[]::new);
+					addToNode(node, vals);
+				} else {
+					addToNode(node, (Long[]) pVal);
+				}
```
david-russo, line 460: Might the remaining cases not also be subject to this issue, e.g. arrays of float or double primitives?

david-russo (commented):
Works for me.

Issue #573, opened 2020-04-27
Replace repeated string literals with constants.


carlwilson:
There are more than 1400 occurrences of this issue.
Why is this an issue?
Checks for multiple occurrences of the same string literal within a single file.

Rationale: Code duplication makes maintenance more difficult, so it can be better to replace the multiple occurrences with a constant.

The Java part of this article: http://rc3.org/2006/10/11/use-contants-instead-of-string-literals/ gives a clear example of how to address these issues.

You can find all occurrences of this issue [here on Codacy](https://app.codacy.com/manual/openpreserve/jhove/issues?bid=3402463&filters=W3siaWQiOiJMYW5ndWFnZSIsInZhbHVlcyI6W119LHsiaWQiOiJDYXRlZ29yeSIsInZhbHVlcyI6W119LHsiaWQiOiJMZXZlbCIsInZhbHVlcyI6W119LHsiaWQiOiJQYXR0ZXJuIiwidmFsdWVzIjpbIjE3NjAiXX0seyJpZCI6IkF1dGhvciIsInZhbHVlcyI6W119XQ==).

The purpose of the current exercise is to tidy up any repetition, sticking to Don't Repeat Yourself principles. For now, it's best to restrict ourselves to eliminating duplication in individual files, and not worry about repetition across files. There are two reasons for this:

- it's hard looking for repetition across files, and Codacy is good at locating it within them; and
- we can ignore issues of scope, that is whether a particular constant should be `public` or `private`, all constants are declared `private`.

These are all constants, so they need to be declared as such with a `final` keyword. Because they're constants there are no threading or instance issues, and it's safe to make them `static` also, so there's a single class constant rather than one for each class instance. This gives us `private static final String` as our standard declaration. Another convention is that class constants have all capital names with `_`s used to separate words, e.g. `MAX_INT` and that all static class constants should be declared at the top of the class.

MartinSpeller:
![](https://github.trello.services/images/mini-trello-icon.png) [Replace repeated string literals with constants. #573 - Assigned to deanforsmith](https://trello.com/c/QypaLciF/70-replace-repeated-string-literals-with-constants-573-assigned-to-deanforsmith)

tomokos2:
Hi, may I help with this issue at all?

carlwilson:
> Hi, may I help with this issue at all?

[...]

PR #586, opened 2020-05-05
Update ConfigHandler.java

archivist-liz:
#573

### DIFF
--- a/jhove-core/src/main/java/edu/harvard/hul/ois/jhove/ConfigHandler.java
+++ b/jhove-core/src/main/java/edu/harvard/hul/ois/jhove/ConfigHandler.java
@@ -371,7 +371,7 @@ public class ConfigHandler
                     throws SAXException, IOException {
         if (systemId.endsWith (configSchemaName)) {
             try {
-                URL resURL = this.getClass().getResource("jhoveConfig.xsd");
+                URL resURL = this.getClass().getResource(configSchemaName);
                 InputStream strm = resURL.openStream ();
                 return new InputSource (strm);
             }
--- a/jhove-core/src/main/java/edu/harvard/hul/ois/jhove/NisoImageMetadata.java
+++ b/jhove-core/src/main/java/edu/harvard/hul/ois/jhove/NisoImageMetadata.java
@@ -23,7 +23,7 @@ public class NisoImageMetadata
     /** private String constants. */
     private static final String YES = "Yes";
     private static final String NO = "No";
-    private static final String COLON = ":";
+    private static final String TIME_STEP = ":";
 
     /******************************************************************
      * PUBLIC CLASS FIELDS.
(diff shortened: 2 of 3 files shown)


jhove-core/src/main/java/edu/harvard/hul/ois/jhove/ConfigHandler.java
```diff
@@ -371,7 +371,7 @@ public InputSource resolveEntity (String publicId, String systemId)
                     throws SAXException, IOException {
         if (systemId.endsWith (configSchemaName)) {
             try {
-                URL resURL = this.getClass().getResource("jhoveConfig.xsd");
+                URL resURL = this.getClass().getResource(configSchemaName);
```
carlwilson, line 374: +1 use of existing constant :)

jhove-core/src/main/java/edu/harvard/hul/ois/jhove/NisoImageMetadata.java
```diff
@@ -20,6 +20,11 @@
 	/** Logger for this class. */
     private static final Logger LOGGER = Logger.getLogger(NisoImageMetadata.class.getCanonicalName());
 
+    /** private String constants. */
+    private static final String YES = "Yes";
+    private static final String NO = "No";
+    private static final String COLON = ":";
```
carlwilson: The definition of the constant `COLON` is a classic "matter of taste" and lacks a definitive answer in lieu of a style guide. Over the years I've taken the definitive stance of wavering on this. This [Stack Exchange question](https://stackoverflow.com/questions/33183525/does-java-define-constants-for-any-characters-such-as-space) states the case against. If it's to be a constant, I'd adopt a name that indicates its use, in this case, `TIME_SEP` perhaps?

carlwilson (changes requested):
See the inline comment, only now do I realise that I've remained true to my wavering in a single review comment, apologies. Either solution is acceptable, ping me in this issue thread or on Slack if you want to discuss anything. Thanks

jhove-core/src/main/java/edu/harvard/hul/ois/jhove/NisoImageMetadata.java
archivist-liz: OK, I got the one for the Niso Image Metadata. I'll update it, even if I get personal sophomoric enjoyment from using COLON as a constant. :) Not sure if I understand what is the issue/if there is an issue with ConfigHandler.java.

jhove-core/src/main/java/edu/harvard/hul/ois/jhove/NisoImageMetadata.java
```diff
@@ -20,6 +20,11 @@
 	/** Logger for this class. */
     private static final Logger LOGGER = Logger.getLogger(NisoImageMetadata.class.getCanonicalName());
 
+    /** private String constants. */
+    private static final String YES = "Yes";
+    private static final String NO = "No";
+    private static final String TIME_STEP = ":";
```
carlwilson: I hate to do this but I meant `TIME_SEP` sorry, short for time separator as that's effectively its use here, sorry I know I'm fussy, it's a general developer character failing and job requirement. :)

carlwilson (changes requested):
I'm sorry but it's good to get some GitHub review practise right? :)

carlwilson (approved):
:) nice one

PR #587, opened 2020-05-05
Udpate PngModule.java 

andreakb:
Addresses issue #573 

Constant FORMAT[0] set as "PNG"; make 2 replacements of literal "PNG" with FORMAT[0]

Literal string "Unused" declared twice in PNG_PROFILES[], delete one "Unused"

Constant PNG_PROFILES[] declared, replace 5 literal strings of png profiles with appropriate PNG_PROFILES[] constant.

### DIFF
--- a/extramodules/it/eng/jhove/module/png/PngModule.java
+++ b/extramodules/it/eng/jhove/module/png/PngModule.java
@@ -239,8 +239,7 @@ public class PngModule extends ModuleBase {
 					   "PNG Truecolor",             // 2
 					   "PNG Indexed",               // 3
 					   "PNG GrayScale with Alpha",  // 4
-					   "Unused",                    // 5
-					   "PNG Truecolor with Alpha"}; // 6
+					   "PNG Truecolor with Alpha"}; // 5
 
     public static final boolean PNG_ENDIANITY=true;
     
@@ -379,7 +378,7 @@ public class PngModule extends ModuleBase {
 										   IdentifierType.URL));
 		_specification.add (doc);
 
-		Signature sig = new InternalSignature ("PNG", SignatureType.MAGIC,
+		Signature sig = new InternalSignature (FORMAT[0], SignatureType.MAGIC,
 											   SignatureUseType.MANDATORY, 0);
 		_signature.add (sig);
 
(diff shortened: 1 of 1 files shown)


extramodules/it/eng/jhove/module/png/PngModule.java
```diff
@@ -379,7 +378,7 @@ public final int parse(final InputStream inputStream,
 										   IdentifierType.URL));
 		_specification.add (doc);
 
-		Signature sig = new InternalSignature ("PNG", SignatureType.MAGIC,
+		Signature sig = new InternalSignature (FORMAT[0], SignatureType.MAGIC,
```
carlwilson, line 381: I'm just becoming aware that this approach does introduce an issue, although it's still an improvement. `FORMAT[0]` isn't as easy to read, but using a single source is a good idea. There's a couple of solutions come to mind but to establish consistency across the codebase I'll need to consider some other files and cases. This is fine for now as it weeds out the repetition.

extramodules/it/eng/jhove/module/png/PngModule.java
```diff
@@ -961,7 +960,7 @@ private final void checkIHDR(final DataInputStream inputStream,
 													colorType + ": " +tmp ));
 
 			}
-			repInfo.setProfile("PNG GrayScale");
+			repInfo.setProfile(PNG_PROFILES[0]);
```
carlwilson, line 963: From here onwards are all improvements regardless as the strings are so long. I suspect that there's a better way of setting these up in the first place using enums but that's beyond the scope of this work and this is the first step.

carlwilson (approved):
There's a couple of points to consider, for me at least, but the approval isn't in any way tentative. Thanks.

extramodules/it/eng/jhove/module/png/PngModule.java
andreakb: Thanks, Carl. Java is still new to me so if you think of a better way to address these categorized strings, let me know. Going forward should I continue to set up constants this way (for instance, there are some repeated error messages in this file that I am thinking of using the enums Error_Message[] for, but  should just use a separate constant for each error message?)

PR #591, opened 2020-05-07
Update WaveModule.java and Parse.Html.java for issue #573

deanforsmith:
I seem to be struggling to submit my pull requests right: the first five commits here were pulled against my repository; the first four are for two files (PngModule.java and ConfigHandler) that in the meantime have been fixed but I haven't been able to remove them. Hopefully, this will properly open a request to review my changes to WaveModule.java and ParseHtml.


jhove-modules/wave-hul/src/main/java/edu/harvard/hul/ois/jhove/module/WaveModule.java
```diff
@@ -254,7 +252,7 @@ public WaveModule() {
 
 		doc = new Document("Specification of the Broadcast Wave Format (BWF)",
 				DocumentType.REPORT);
-		doc.setIdentifier(new Identifier("EBU Technical Specification 3285",
+		doc.setIdentifier(new Identifier(FORMAT[2],
```
david-russo: Unfortunately we can't rely on the order of the FORMAT array to stay the same in the future. If a future programmer adds another format value to anywhere except the end of the array, the value which this index references (and any following it) may change to an incorrect value. Let me demonstrate. If we had an array as follows:
```
String[] FORMATS = { "WAVE", "Broadcast Wave Format", "BWF" }
```
And had code expecting the positions of the array content to stay unchanged:
```
"The " + FORMAT[1] + " is often abbreviated to " + FORMAT[2]
```
While initially, the above would output "The Broadcast Wave Format is often abbreviated to BWF", after the addition of a missing format identifier to the array, like so:
```
String[] FORMATS = { "WAVE", "Waveform Audio", "Broadcast Wave Format", "BWF" }
```
The sentence would then change to "The Waveform Audio is often abbreviated to Broadcast Wave Format". These kinds of unintended changes can be hard to spot, especially with long-lived code maintained by multiple developers, so we guard against them by avoiding references into arrays using specific array indices where we can.

One of the exceptions to this you might see around the codebase are references to the first item in the `FORMATS` array (`FORMAT[0]`), which is _assumed_ to be populated with at least one value, the first of which is also assumed to be the preferred human-readable label for the main format, for use in reports. Assumptions in code are never ideal, so there's probably another task there for the future.

For everything else, including this line, you're better off creating new `static final` variables to reference here and in the `FORMAT` array, e.g. `EBU_TECH_SPEC_3285`.

jhove-modules/wave-hul/src/main/java/edu/harvard/hul/ois/jhove/module/WaveModule.java
```diff
@@ -297,15 +295,15 @@ public WaveModule() {
...
+		sig = new InternalSignature(RIFF_SIGNATURE, SignatureType.MAGIC,
 				SignatureUseType.MANDATORY_IF_APPLICABLE, 0);
 		_signature.add(sig);
 
 		sig = new InternalSignature("RF64", SignatureType.MAGIC,
 				SignatureUseType.MANDATORY_IF_APPLICABLE, 0);
 		_signature.add(sig);
 
-		sig = new InternalSignature("WAVE", SignatureType.MAGIC,
+		sig = new InternalSignature(FORMATS[0], SignatureType.MAGIC,
```
david-russo: This change (and the following on line 364) are not quite like the others, though they understandably appear identical, they in fact have different meanings and uses.

The values in the `FORMAT` array are intended as human-readable strings used to report all the formats this module can identify. The `"WAVE"` value here is instead an identifier intended to match against the same identifier expected in a WAVE file's file format structure, like the `RIFF_SIGNATURE` value above it. So while the `"WAVE"` value in the `FORMAT` array could change in the future to be `"Wave"` for stylistic reasons, the `"WAVE"` value here and on line 364 must remain as `"WAVE"` to match the expected format identifier.

To guard against such a change, we should consolidate repeated string values based on meaning and usage, instead of only the value. To do that here, we could move this instance of `"WAVE"` to a new variable called `WAVE_FORM_TYPE`, and reference it here and on line 364.

extramodules/it/eng/jhove/module/png/PngModule.java
```diff
@@ -370,7 +370,7 @@ public final int parse(final InputStream inputStream,
 										   IdentifierType.URL));
 		_specification.add (doc);
 
-		Signature sig = new InternalSignature ("PNG", SignatureType.MAGIC,
+		Signature sig = new InternalSignature (COVERAGE, SignatureType.MAGIC,
```
david-russo: See [my note](https://github.com/openpreserve/jhove/pull/591#discussion_r422939620) on WaveModule.java, line 306. This is another internal identifier with a separate meaning and usage to the value in `COVERAGE`.

david-russo (changes requested):
Good work on the HTML parser and ConfigHandler. There are just a few very unobvious complexities to the changes in the PNG and Wave modules that need some alterations.

jhove-modules/wave-hul/src/main/java/edu/harvard/hul/ois/jhove/module/WaveModule.java
deanforsmith: Thanks, @david-russo , this is very clear and helpful. As a learning exercise, if I may ask (just to be sure I fully understand) - the idea would be to leave the array as-is to keep a human-readable list of formats but add additional `static final` variables to eliminate the repeated string literals? And by extension, while `FORMAT[0]` is less likely to be problematic, it could be, and in any case if only for stylistic consistency should probably also have a separate `static final`?

jhove-modules/wave-hul/src/main/java/edu/harvard/hul/ois/jhove/module/WaveModule.java
deanforsmith: Now that you've pointed this out, I will be on alert to spot such differences going forward, but it's possible that these less-obvious uses may be beyond my current Java skill. Thanks again, @david-russo : I appreciate your time walking me through the code base.

jhove-modules/wave-hul/src/main/java/edu/harvard/hul/ois/jhove/module/WaveModule.java
david-russo: Spot on. So using the above code as an example, I would create a new variable and place it in the code wherever that string is used with the same meaning (in the case of the tech. spec. name, it's all just for human-readable documentation):

```java
private static final EBU_TECH_SPEC_3285 = "EBU Technical Specification 3285";
...
private static final String[] FORMATS = { "WAVE", ..., EBU_TECH_SPEC_3285, ...
...
doc.setIdentifier(new Identifier(EBU_TECH_SPEC_3285, ...
```

And yes, the fix for `FORMAT[0]` usage is probably just replacing it with references to appropriate `static final` variables.

jhove-modules/wave-hul/src/main/java/edu/harvard/hul/ois/jhove/module/WaveModule.java
david-russo: That's OK — that's what these reviews are for. Just flag up any uncertainties you have and we'll have a look at them.

MartinSpeller:
[Replace repeated string literals with constants. #573 - Assigned to deanforsmith](https://trello.com/c/QypaLciF/70-replace-repeated-string-literals-with-constants-573-assigned-to-deanforsmith)

carlwilson (approved):

david-russo:
Thanks for your fixes, Dean. There's just [one remaining above](https://github.com/openpreserve/jhove/pull/591#discussion_r422950507) in PngModule before we can merge this in. The internal signature declaration should use a separate variable from `COVERAGE`, something like `PNG_SIG`.

david-russo (changes requested):
Thanks for your fixes, Dean. There's just [one remaining](https://github.com/openpreserve/jhove/pull/591/files#r422950507) in PngModule before we can merge this in. The internal signature declaration should use a separate variable from `COVERAGE`, something like `PNG_SIG`.

carlwilson:
All changes here are merged now I believe, closing.

PR #596, opened 2020-05-11
PDF-HUL-117: rotation property not a Simple object

samalloing:
Hi @carlwilson,
I was working on the error PDF-HUL-117 and I tested the provided Example. I noticed the rotation property was null and that this should process properly. The error what occurred was a casting error. The problem is that the object returned for DICT_KEY_ROTATE was a PdfIndirectObj and not a PdfSimpleObject. So I added code to make the PdfIndirectObj a PdfSimpleObject. Now the Example is processed without any problems and is the file valid and well-formed.
I expect the same thing can happen for the other properties. If you approve this, I change the other properties as well.

I'm sorry if this is the wrong way of doing things, but I didn't succeed to do cherry-picking with one commit. I hope this doesn't messes up things

Sam

### DIFF
--- a/extramodules/it/eng/jhove/module/odf/OdfModule.java
+++ b/extramodules/it/eng/jhove/module/odf/OdfModule.java
@@ -102,6 +102,10 @@ public class OdfModule extends ModuleBase implements Module {
 											  "Open Document Format Global Text Document",
 											  "Open Document Format HTML Text Document Template"};
 
+    private static final String TEMP_ODF = "odfTemp";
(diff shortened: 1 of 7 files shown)


carlwilson (approved):
OK, I get it. It's a case of detecting the reference case (Indirect object) and resolving. If there are other examples then fixes would be welcome thanks @samalloing

carlwilson:
I see that this has made its way into the code base via another route. That's fine, and the build is all good. Thanks, @samalloing.

PR #597, opened 2020-05-13
Fixes #375: Handle class cast from indirect object to simple object

### LINKED ISSUE #375: Unexpected exception lava.lang.ClassCastException in PDF-hul 1.11 JOHVE 1.20.1
## Dev Effort
2D

## Description
[F-03-B-823.pdf](https://github.com/openpreserve/jhove/files/2680299/F-03-B-823.pdf)

We have a large number of files all from the same collection that cause the same unexpected exception in JHOVE. I'm attaching one of them and happy to provide more if needed.

karenhanson:
This fix for issue #375 repairs a ClassCastException relating to casting from an indirect object to a simple object. This seems to be a common conversion throughout the PdfModule that has been handled using the `resolveIndirectObject` method. This fix simply passes the object causing the problem into that method. I didn't write a test since this seems to be a repeated feature of the code, and the test would involve getting permission to use the file supplied with the issue. Please let me know if I should write it anyway.

It looks like PR #596 may be responding to a similar issue.

### DIFF
--- a/jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/PdfModule.java
+++ b/jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/PdfModule.java
@@ -2302,8 +2302,8 @@ public class PdfModule extends ModuleBase {
 											if (parms != null) {
 												PdfSimpleObject kobj = null;
 												if (parms instanceof PdfDictionary) {
-													kobj = (PdfSimpleObject) ((PdfDictionary) parms)
-															.get(DICT_KEY_K);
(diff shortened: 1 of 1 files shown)


carlwilson (approved):
So you're right that #596 covers similar ground. I think that there's likely a few of these in the code. Nice fix though.

MartinSpeller:
[Unexpected exception lava.lang.ClassCastException in PDF-hul 1.11 JOHVE 1.20.1 #375 - Assigned to Karen Hansen](https://trello.com/c/KKmoAcRk/57-unexpected-exception-lavalangclasscastexception-in-pdf-hul-111-johve-1201-375-assigned-to-karen-hansen)

PR #607, opened 2020-05-18
missing Encryption Algorithm

samalloing:
Hi @carlwilson 
According to https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/adobe_supplement_iso32000.pdf is AESv3 a new Encryption Algorithm. It is added as an extension from Extension level 3 onward.
An example file is: https://github.com/qpdf/qpdf/blob/master/qpdf/qtest/qpdf/copied-positive-P.pdf

Now PDF-HUL-93 is the error, but with this patch this problem is solved.


samalloing:
added extra code for parsing the new parameters associated with this new algorithm. So I close this pull request. The new pull request is #621

PR #608, opened 2020-05-20
Update ProgressWindow.java for Issue #573

### LINKED ISSUE #573: Replace repeated string literals with constants.

andreakb:
Addresses Issue #573

Created and inserted constants for repeated strings.


MartinSpeller:
[Replace repeated string literals with constants. #573 - Assigned to deanforsmith](https://trello.com/c/QypaLciF/70-replace-repeated-string-literals-with-constants-573-assigned-to-deanforsmith)

carlwilson (approved):
Lovely :)

PR #615, opened 2020-05-29
Replace += with StringBuilder for whitespace to speed up Tokenizer

### LINKED ISSUE #614: PDF with a large amount of whitespace takes days to process
We have a PDF that is not well formed. It has ~100MB of whitespace in the middle of it. The file gets stuck in a loop in `Tokenizer.getNext (long max)` - the loop is processing 1 whitespace at a time and adding it to _wsString. I can see the offset is incrementing, but it is running for several days and only processed 1/3 of the file. I note that while the token buffer is using a StringBuilder, the whitespace String is using `+=`. Not sure if this is causing the slowness. Unfortunately I can't share the example PDF. 

Note: I have a PR coming that will test the theory!

karenhanson:
This is a potential fix for issue #614 - whitespace is processed very slowly in the PDF Tokenizer. This became a problem when we had some files with >100MB of whitespace. Though the PDFs were not valid, JHOVE ran for days without getting to the end of the whitespace on a 160MB file.  The fix uses a `StringBuilder` instead of doing `+=` on a String. It is now suspiciously fast, only taking a few seconds - which makes me wonder if I've missed something in the logic!

Testing (_edited_): I have created a test file that can be used to replicate/test this issue and attached it to issue #614. I confirmed that this change reduces the processing time to seconds on both the original that we had the problem with (which I can't share) and my manufactured test file. I've submitted this change without a test for now - please let me know if I should add a test using my test file or if it needs to be added to a test corpus elsewhere.
 
_Note: The issue that this relates to is newly logged and hasn't been evaluated yet - please let me know if it would be best to withdraw this PR until the issue is reviewed. My apologies if I've done things out of order!_


MartinSpeller:
[PDF with a large amount of whitespace takes days to process #614 - Assigned Karen Hanson](https://trello.com/c/IGbn22g4/80-pdf-with-a-large-amount-of-whitespace-takes-days-to-process-614-assigned-karen-hanson)

carlwilson (approved):

PR #625, opened 2020-06-12
FEAT: Piccolo Proof of Concept

DinoAGW:
* First complete try to use Picolo as a config objectl and
* fixed faulty package-info stuff.


carlwilson:
@david-russo and @tledoux I've asked you both to take a look at this as I know you'd both expressed opinions. I'm going to write some tests for illegal parameter combos and the like and add param descriptions. It also seems sensible to factor out the description strings as resources. That aside, it all looks good.

david-russo (commented):
Assuming all the TBD are addressed, and it supports at least what the current one does, the use of picocli already looks like a vast improvement to me. I don't see any problems with the logic either, it only seems to need some comment clean up and option descriptions.

Passing the old `-P` or `-p` options will now probably cause the CLI to exit with an 'unknown option' message of some kind (I haven't actually run the code). Those options have had no effect for a while though, so this is probably as good a time as any to alert people to that.

PR #631, opened 2020-06-27
Java source normalization (fixes #516)

### LINKED ISSUE #516: Standardize source code formatting
Currently, the source code is using very different conventions of code formatting (tabs or spaces for indentation,  curly braces with LF before or after, separation of function name and parameters, ...). This make the commits quite hard to read and generates useless conflicts.

The idea is to define one standard for the whole project and force the commits to respect it.

A way to go could be to use [google-java-format](https://github.com/google/google-java-format) which is provided in Java (so works on every computer that tries to dev on Jhove), works with maven, provides plug-ins for the main IDE and most importantly can be define as a pre-commit hook for git which will make sure every file is indeed formatted as the others.

The initial commit (reformatting everything will be painfull) but in the long run it should make life easier...

svanteschubert:
I have fixed the same problem today for the ODF Toolkit: https://github.com/tdf/odftoolkit/issues/50

Using the Google Java Formatter:
https://github.com/google/google-java-format 

Capsulated by the following Maven plugin:
https://github.com/Cosium/maven-git-code-format
adding a pre-commit-hook for checking the Java code.

You might reproduce the patch by:

1. Add the plugin to the pom.xml (as in the first commit)
2. Start manual formatting on the sources by command line via:
`mvn -X git-code-format:format-code -Dgcf.globPattern=**/*`
There will be half a dozen problems with comments of unused imports within the Java imports, which breaks the formatting, but after the manual fixes, you will come to my patch.

NOTE: The ODF Toolkit is using in addition to the above Java formatter a whitespace formatter for the text files:
https://ec4j.github.io/editorconfig-maven-plugin/

See in https://github.com/tdf/odftoolkit/blob/master/pom.xml#L257
and its configs:
https://github.com/tdf/odftoolkit/blob/master/.editorconfig


svanteschubert:
Please note, that this normalization step causes a lot of noise to prevent noise in the future.
You might want to consider to merge other pull requests prior, or the pull requests have to do the normalization themselves (fixes all the import problems). Perhaps those import problems should be separated first? Your call :-)

carlwilson:
Now we're back up and running properly, I'm interested. I'll look to reproduce this once 1.26 is out of the door.

Issue #645, opened 2020-09-10
StackOverflowError in with 1.24.1 in PDF-hul


asciim0:
The attached file throws an error in JHOVE 1.24.1 and can't be validated

> Exception in thread "main" java.lang.StackOverflowError
> 	at edu.harvard.hul.ois.jhove.module.pdf.Tokenizer.readChar1(Tokenizer.java:704)
> 	at edu.harvard.hul.ois.jhove.module.pdf.Literal.readBackslashSequence(Literal.java:609)
> 	at edu.harvard.hul.ois.jhove.module.pdf.Literal.processLiteral(Literal.java:289)
(...)

[1027509533.pdf](https://github.com/openpreserve/jhove/files/5202072/1027509533.pdf)

Issue #646, opened 2020-09-10
TimeOut / stuck in loop (?) - 1.24.1 PDF-hul for attached file


asciim0:
Using both UI and CLI (on Windows environment, tested on different machines) the attached file nevers seems to reach a result for validation.

[897714407.pdf](https://github.com/openpreserve/jhove/files/5202088/897714407.pdf)

rosetta-development:
We encounter the same issue with our pdf file.
Thread is stuck on : 
`edu.harvard.hul.ois.jhove.Property.getByName(Property.java:159)`

karenhanson:
I think I have this same issue  - so wanted to share what I found while trying to troubleshoot in case it adds useful context. Looks like #645 may be the same thing too. In my case the following lines in the PDF are causing an infinite loop (using v1.24):
```
464 0 obj
<</Dest(þÿ s e c S 0 0 4)/Next 463 0 R/Parent 399 0 R/Prev 465 0 R/Title(þÿ P a r t A\t & \t .)>>
endobj
465 0 obj
<</Dest(þÿ s e c S 0 0 3)/Next 464 0 R/Parent 399 0 R/Prev 466 0 R/Title(þÿ P a r t B)>>
endobj
```
In the `Tokenizer` it seems that the backslash in the title causes it to go into `Literal.readBackslashSequence()` where it fails to see the `>>` that should end the entry. It then proceeds to read in the following row as part of the same object, which sets the `Next` value back to 464 and causes an infinite loop where it keeps reloading the same garbled object. In my case the PDF gets stuck [in this loop](https://github.com/openpreserve/jhove/blob/47f077fcb09d3dd5720aeeb4abf27de5dac48d19/jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/AProfile.java#L835-L841).

I note that the PDF attached to this issue and the one attached to #645 both have backslashes in the Title property of a dictionary entry. I tested the fix in PR #652 and it works for my issue too.

Issue #473, opened 2019-09-04
validation process with specific pdf stuck in infinite loop


digitalpreservationhero:
Validation process with the attached pdf ran for over 12 hours using JHOVE 1.22 pdf-hul before I aborted the process. I assume, there's something in the pdf structure which causes JHOVE to stuck in an infinite loop. 
[infinite_loop.pdf](https://github.com/openpreserve/jhove/files/3573545/infinite_loop.pdf)

carlwilson:
Hi @digitalpreservationhero, thanks for raising this. I've just tried this on the new v1.24 release candidate and it no longer runs in an infinite loop, or at least not for me. It now fails with a `StackOverflowError` and the offending (recursive) method appears to be `edu.harvard.hul.ois.jhove.module.pdf.AProfile.checkOutlineItem(AProfile.java:864)`. This isn't going to be fixed for the upcoming release but I'll prioritise it for the next release.

MartinSpeller:
![](https://github.trello.services/images/mini-trello-icon.png) [validation process with specific pdf stuck in infinite loop #473 - Assigned to TBA](https://trello.com/c/STcuL1fi/46-validation-process-with-specific-pdf-stuck-in-infinite-loop-473-assigned-to-tba)

ross-spencer:
I was interested to see if there was any relation to https://github.com/openpreserve/jhove/issues/115

In terms of logic, I'd need to look into this in more detail to know if there is a relation. In terms of behavior then it looks like JHove 1.25 integration branch (and 1.24 did) exits with a stack overflow and stack trace now and no longer enters an infinite loop:
```
Exception in thread "main" java.lang.StackOverflowError
	at java.lang.StringCoding$StringEncoder.encode(StringCoding.java:304)
	at java.lang.StringCoding.encode(StringCoding.java:344)
	at java.lang.StringCoding.encode(StringCoding.java:387)
	at java.lang.String.getBytes(String.java:958)
	at edu.harvard.hul.ois.jhove.module.pdf.Name.isPdfACompliant(Name.java:23)
	at edu.harvard.hul.ois.jhove.module.pdf.Tokenizer.getNext(Tokenizer.java:424)
	at edu.harvard.hul.ois.jhove.module.pdf.Parser.getNext(Parser.java:94)
	at edu.harvard.hul.ois.jhove.module.pdf.Parser.getNext(Parser.java:83)
[...]

Issue #306, opened 2018-01-17
JHOVE hangs for PDF 1.7 file


rolandsuri:
## Dev Effort
0.5D - investigation 

## Description
JHOVE gets stuck with a complex, but most likely valid, PDF 1.7 File (PDF attached). I used JHOVE Version 1.18.1 with PDF-hul 1.10. JHOVE reported an Exception in thread "Thread-4" Java.lang.StackOverflowError (screenshot shown below). I know that this module does not cover PDF 1.7, but still, it should react more gracefully.

[The role of low-density receptor-related protein 1 (LRP1) as a competitive substrate of the amyloid precursor protein (APP) for BACE1.pdf]
(https://github.com/openpreserve/jhove/files/1638584/The.role.of.low-density.receptor-related.protein.1.LRP1.as.a.competitive.substrate.of.the.amyloid.precursor.protein.APP.for.BACE1.pdf)
![java-lang-stackoverflowerror](https://user-images.githubusercontent.com/17157009/35038226-0ed537ec-fb72-11e7-9571-762c3160b703.JPG)

ghost:
To be tested in 1.22 in hack week

MartinSpeller:
![](https://github.trello.services/images/mini-trello-icon.png) [JHOVE hangs for PDF 1.7 file #306 - Assigned to TBA](https://trello.com/c/RdFsLEYD/29-jhove-hangs-for-pdf-17-file-306-assigned-to-tba)

PR #652, opened 2020-11-20
Fix issues with backslash sequence

MaximPlusov:
Fix issues #645 and #646

### DIFF
--- a/jhove-bbt/scripts/create-1.25-target.sh
+++ b/jhove-bbt/scripts/create-1.25-target.sh
@@ -98,3 +98,18 @@ if [[ -f "${candidateRoot}/regression/modules/PNG-gdm/issue_148.png.jhove.xml" ]
 	cp "${candidateRoot}/regression/modules/PNG-gdm/issue_148.png.jhove.xml" "${targetRoot}/regression/modules/PNG-gdm/issue_148.png.jhove.xml"
 fi
 
+# Patch the offset in errors/modules/PDF-hul/pdf-hul-76-372051162.pdf.jhove.xml
+# fixed by https://github.com/openpreserve/jhove/pull/652
+echo " - ISSUE:645 & ISSUE:646 PDF result patch 1."
+find "${targetRoot}" -type f -name "pdf-hul-76-372051162.pdf.jhove.xml" -exec sed -i 's/^   <message offset="268334" severity="error" id="PDF-HUL-66">Lexical error<\/message>$/   <message offset="268333" severity="error" id="PDF-HUL-66">Lexical error<\/message>/' {} \;
+# Copy the full result of regression/modules/PDF-hul/issue_645.pdf and
+# regression/modules/PDF-hul/issue_646.pdf
+# fixed by: https://github.com/openpreserve/jhove/pull/652
+if [[ -f "${candidateRoot}/regression/modules/PDF-hul/issue_645.pdf.jhove.xml" ]]; then
+	echo " - ISSUE:645 & ISSUE:646 PDF result patch 2."
+	cp "${candidateRoot}/regression/modules/PDF-hul/issue_645.pdf.jhove.xml" "${targetRoot}/regression/modules/PDF-hul/issue_645.pdf.jhove.xml"
+fi
+if [[ -f "${candidateRoot}/regression/modules/PDF-hul/issue_646.pdf.jhove.xml" ]]; then
+	echo " - ISSUE:645 & ISSUE:646 PDF result patch 2."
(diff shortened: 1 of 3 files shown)


carlwilson (approved):
So focussing the backslash check on the second UTF byte is the main change here? Not sure what the previous link to uftch != ESC was, or put another way is there any reason to worry that  it's now theoretically possible for both readBackslashSequence and readUTFLanguageCode to be invoked?

karenhanson:
Thanks for your work on this! We tested it on a number of PDFs that had this stack overflow issue and it fixed the majority of them. I found a couple that still have a problem. I only had time to dig into one of them, but the problem seems to be related to some variation in the characters before and after the backslash. For example, in this case it looks like the PDF editor that produced the following sequence inserted the backslash+CR as single-bytes but split some double-byte characters in the process.  
 
image
(Sorry for the horrible notepad++ screenshot, hopefully it illustrates the issue).

The backslash section of Literal.java has a lot of comments - seems like it's really difficult to make clear assumptions about how the backslash is added.

I'm still learning this stuff, so I'm not sure what makes sense, but I wonder if this line needs to be conditional on what the next character is -
https://github.com/openpreserve/jhove/blob/47f077fcb09d3dd5720aeeb4abf27de5dac48d19/jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/Literal.java#L296
In some cases perhaps it should hop back to UTF16_1? 

Let me know if you think any of this warrants a separate issue rather than something that might be part of this PR and I will log accordingly. I can also help fix code if I know what it _should_ do.

karenhanson:
Thinking about this some more - I realized reading the next character wouldn't be helpful. If there isn't logic that could detect all variations of a badly placed backslash, maybe the solution is something to manage the side effect. The key side effect is that Literal.processLiteral reads in the close parenthesis as the second byte of a UTF16 character - my case "e)" - causing it not to act on the close parenthesis and end the Literal. So processLiteral continues to read past the end of the object to the next one. In my case, it reads in the Next property of the following object which causes a recursive loop. Maybe detecting recursion and breaking out of it with a warning is a reasonable precaution to allow for the encoding issues we can't predict? If that makes sense it's probably a separate issue.

bdoubrov:
@karenhanson the syntax for PDF string objects, including the use of escape sequences and UTF16 (PDF 2.0 also introduces UTF8 support) is well documented in PDF 1.7 specification, clause 7.3.4. I don't remember any ambiguities in this part of the spec. 

In particular, the implementation logic should first real all literal string content between parentheses processing only escape sequences (bot not UTF16 encoding yet) and resulting in the byte array. This byte array then will be passed to the UTF16 encoding algorithm to turn it into the Unicode string.

If I understand correctly your example, the issue will show up not in the parsing of the literal string as a byte array, but when interpreting it as a UTF16-encoded Unicode string. 

We'll check if the current JHOVE code does exactly this and suggest fixes, if not. It would also be helpful if you could attach the PDF document with this /Title string issue.

karenhanson:
Thanks @bdoubrov, I really appreciate that context! That makes sense - I can step through the code to see what the glitch is but I don't know much about the PDF standard and use of escape characters. I believe `Literal.processLiteral` is just working through the file byte by byte from a particular offset, so it could theoretically hit EOF if it doesn't detect an appropriate end Literal character. In other words, I believe it _may_ be doing the two steps you mentioned simultaneously in one loop.... reading byte by byte to look for escape characters and interpreting as unicode while reading.

I can't share the PDF with the error unfortunately, but I looked at the history of this issue and I think there are at least 5 open tickets related to it (#277, #306, #473, #645, #646). I tested all of the PDFs I could find that were attached to these to see if any are not yet fixed by this PR. 4 of these tickets appear to be fixed (which matches my experience that this PR resolves the majority of these errors), but the [file attached](https://github.com/openpreserve/jhove/files/1638584/The.role.of.low-density.receptor-related.protein.1.LRP1.as.a.competitive.substrate.of.the.amyloid.precursor.protein.APP.for.BACE1.pdf) to #306 still errors even with this fix. Looking at the objs in that PDF I would guess this is because there is a Title literal that uses UTF16 encoding with single-byte backslash + CR... as with my example, the following line of code probably causes byte 1 on the next line of the Title after the backslash and CR to be read as byte 2:
https://github.com/openpreserve/jhove/blob/47f077fcb09d3dd5720aeeb4abf27de5dac48d19/jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/Literal.java#L296
That means as it loops byte 2, 1, 2, 1... the nulls and eventually the close parenthesis become the second byte in this sequence... close parenthesis is not managed if it is on byte 2 of UTF16 and so it goes past the end of the line.

On the plus side, I think the current proposed PR allows the PDFs attached to #277 and #473 to be processed in addition to those attached to the two issues already tagged.

Issue #572, opened 2020-04-27
Junit test errors in JSONHandlerTest


ross-spencer:
Trying to build from source to prepare for the upcoming week I'm seeing the following failures in the JSON handler tests. It seems that the expected date is not being matched and therefore replaced with `DUMMY` date. 
```
Failed tests: 
  JsonHandlerTest.testShow:167 test expected:<...1","executionTime":"[DUMMY]"}}> but was:<...1","executionTime":"[2020-04-26T20:29:43-04:00]"}}>
  JsonHandlerTest.testShowApp:187 expected:<...1","executionTime":"[DUMMY]","app":{"api":{"ver...> but was:<...1","executionTime":"[2020-04-26T20:29:43-04:00]","app":{"api":{"ver...>
  JsonHandlerTest.testShowModule:230 expected:<...1","executionTime":"[DUMMY]","module":{"name":"...> but was:<...1","executionTime":"[2020-04-26T20:29:43-04:00]","module":{"name":"...>
  JsonHandlerTest.testShowOutputHandler:208 expected:<...1","executionTime":"[DUMMY]","handler":{"name":...> but was:<...1","executionTime":"[2020-04-26T20:29:43-04:00]","handler":{"name":...>
  JsonHandlerTest.testShowRepInfo:259 expected:<...1","executionTime":"[DUMMY]","repInfo":{"uri":"...> but was:<...1","executionTime":"[2020-04-26T20:29:43-04:00]","repInfo":{"uri":"...>
```
It looks like the date-time replacement string needs to include timezone differences, extending the regex to include plus/minus timezone differences (optionally) might do the trick:
```
-       private static final String TIME_PATTERN = "\"[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}(\\+[0-9]{2}:[0-9]{2})?\"";
+       private static final String TIME_PATTERN = "\"[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}(\\+[0-9]{2}:[0-9]{2})?((-|\\+)[0-9]{2}:[0-9]{2})?\"";
```

The addition is simply: `((-|\\+)[0-9]{2}:[0-9]{2})?`

We can tests on regex101: [here](https://regex101.com/r/f4J2X0/1)

E.g. `2020-04-26T20:29:43-04:00`

[...]

PR #684, opened 2021-07-06
fix time replacement regex in test

pwinckles:
Fixes a regex that's used to replace times in tests to account for timezones that have negative offsets.


ross-spencer:
@pwinckles should be connected to https://github.com/openpreserve/jhove/issues/572? (PS. this is really helpful thx!)

pwinckles:
> 
> 
> @pwinckles should be connected to #572? (PS. this is really helpful thx!)

Yep, I hadn't noticed that ticket

carlwilson:
I'm finally getting round to a merge up for an imminent 1.26 release. Thanks for the work here @pwinckles and @ross-spencer

PR #685, opened 2021-07-06
only validate xml when schema locations are known

### LINKED ISSUE #680: XML should not be validated when no schema provided
The [documentation](https://jhove.openpreservation.org/modules/xml/#validity) states:

> Note that the concept of validity applies only to XML files that explicitly reference a DTD or XML Schema. JHOVE can determine if either of these conditions are met and if so, it will invoke automatically the SAX2 parser in a validating mode. Otherwise, the parser is invoked in a manner that only checks for well-formedness.

However, this is not what the code actually does. Instead, it will validate any XML document that contains a namespace regardless of whether or not there is a schema or dtd.

Here is a example that demonstrates the problem:

```xml
<example xmlns="http://example.com">
  <text>This is an example</text>
</example>
```

```shell
❯ ./jhove -m XML-hul example.xml 
Jhove (Rel. 1.24.1, 2020-03-16)
 Date: 2021-06-24 07:41:51 CDT
 RepresentationInformation: example.xml
  ReportingModule: XML-hul, Rel. 1.5.1 (2019-04-17)
  LastModified: 2021-06-24 07:41:36 CDT
  Size: 82
  Format: XML
  Version: 1.0
  Status: Well-Formed, but not valid
  SignatureMatches:
   XML-hul
  ErrorMessage: SaxParseException: cvc-elt.1: Cannot find the declaration of element 'example'.: Line = 1, Column = 37
   ID: XML-HUL-1
  MIMEtype: text/xml
  XMLMetadata: 
   Parser: org.apache.xerces.jaxp.SAXParserImpl$JAXPSAXParser
   Encoding: UTF-8
   Schemas: 
    Schema: 
     NamespaceURI: http://example.com
     SchemaLocation: 
   Root: example
   Namespaces: 
    Namespace: 
     Prefix: 
     URI: http://example.com

```

The expected behavior is that JHOVE only checks for well-formedness and does **not** validate the xml.

pwinckles:
Resolves https://github.com/openpreserve/jhove/issues/680

Changes the XML module to _not_ attempt to validate XML that does not have a known schema location.

### DIFF
--- a/jhove-modules/xml-hul/src/main/java/edu/harvard/hul/ois/jhove/module/XmlModule.java
+++ b/jhove-modules/xml-hul/src/main/java/edu/harvard/hul/ois/jhove/module/XmlModule.java
@@ -450,6 +450,7 @@ public class XmlModule extends ModuleBase {
 		// It should be noted that later on, when we check the namespace
 		// of the root element, if it has an associated URI, that will
 		// be used instead.
+		boolean hasRootSchema = false;
 		if (!schemaList.isEmpty()) {
 			SchemaInfo schItems = schemaList.get(0);
 			if (isNotEmpty(schItems.namespaceURI)) {
@@ -457,13 +458,18 @@ public class XmlModule extends ModuleBase {
 			} else if (isNotEmpty(schItems.location)) {
 				_textMD.setMarkup_language(schItems.location);
 			}
+
+			if (isNotEmpty(schItems.location)
+					|| (isNotEmpty(schItems.namespaceURI) && _localSchemas.containsKey(schItems.namespaceURI))) {
+				hasRootSchema = true;
+			}
 		} else if (isNotEmpty(dtdURI)) {
 			_textMD.setMarkup_language(dtdURI);
+			hasRootSchema = true;
 		}
 
 		if (parseIndex == 0) {
-			if ((handler.getDTDURI() != null || !schemaList.isEmpty())
-					&& canValidate) {
+			if (canValidate && hasRootSchema) {
 				return 1;
 			}
 			info.setValid(RepInfo.UNDETERMINED);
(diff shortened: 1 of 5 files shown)


pwinckles:
@carlwilson I rebased this PR, any chance it can be included in the 1.26 release? Thanks!

carlwilson:
@pwinckles given I've bodged the final build packaging there's every chance. Will test and include in an imminent RC2

carlwilson:
There's an issue with the Java 8 build and test here, a different XML implementation I'd guess. Here again: https://github.com/openpreserve/jhove/pull/740 for the merge PR. I can take a brief look to see if I can fix it tomorrow but will be a time-boxed inspection. If I can't resolve by tomorrow AM US time I'll release without this patch sorry.

pwinckles:
@carlwilson I debugged the issue. The problem is that it can't load the DC schema at http://dublincore.org/schemas/xmls/simpledc20021212.xsd. It gets a 403. Here's a reproduction of the issues:

```java
        URLConnection c = new URL("http://dublincore.org/schemas/xmls/simpledc20021212.xsd").openConnection();
        InputStream s = c.getInputStream();
        s.close();
```

Dublincore had a DDoS attack a few months back and I think put some precautions in place. The above code will work if you add a User-Agent to the connection. Unfortunately, Java 8 does not do this. I assume releases since then include an agent, which is why it works on 11+.

I can try to come up with a different example for the test case.

pwinckles:
Okay, I think everything should be good now

carlwilson:
Sorry, a time difference thing. I was off to bed when I sent the last message. This all seems good now and will be in the RC out for your morning I think. Thanks very much for the work and all, there will be more TLC for JHOVE this year.

pwinckles:
Thank you!

PR #705, opened 2022-02-02
add support for specifying xml schema urls

pwinckles:
Previously, XML schemas could only be set by specifying local files on disk in the jhove config. Now, the same config accepts URLs.


pwinckles:
Whoops. Opened this in the wrong place.

PR #718, opened 2022-04-06
FEAT: Basic Github Actions build

carlwilson:
- checkout project;
- run mvn build and verify;
- run the JHOVE integration tests; and
- upload static QA to Codacy.


CGarces:
Hi @carlwilson 

Matrix is declared on the build job
```
    strategy:
      matrix:
        java-version: [8, 11]
```

But  `if: matrix.java-version == 8` is used on the coverage section. I think that this will never work.

PR #731, opened 2022-04-13
FIX: ConfigHander tests for other locales.

carlwilson:
- fixed issue with `ConfigHandlerTest` when called from non-English locales;
- fixed the rogue package declarations in `jhove-ext-modules`


carlwilson:
OK @tledoux, I admit it, your idea was better, GH actions liked my fix (it worked back here as well as I swapped locales) but Travis wouldn't have it.

tledoux:
Hi @carlwilson , cool if it solves the problem in all environments.
By the way, since the release seems to be in a good shape, would it be possible to regenerate ALL the jars (even from untouched modules) so that they are all at the same level of signature or lack of it ? 
It will make live easier for the ones integrating Jhove in their products. Thanks again.

Issue #358, opened 2018-09-05
PDF: Parentheses not handled properly in document information dictionary


T3rj3:
## Dev Effort
0.5D

## Description

[This PDF-file](https://github.com/openpreserve/jhove/files/2351907/sample.pdf) fails validation with the following error:
>Jhove (Rel. 1.20.1, 2018-03-29)
 Date: 2018-09-05 10:07:25 CEST
 RepresentationInformation: /home/sample.pdf
  ReportingModule: PDF-hul, Rel. 1.11 (2018-03-29)
  LastModified: 2018-08-30 10:10:10 CEST
  Size: 2736147
  Format: PDF
  Status: Not well-formed
  SignatureMatches:
   PDF-hul
  ErrorMessage: Lexical error
   Offset: 2731501
  MIMEtype: application/pdf

This happens because the file's document information dictionary contains two entries with parentheses in them:

>/CreationDate (20180606085937+01'00')
  /Creator (QuarkXPress(R) 8.16r2)
  /ModDate (20180606085937+01'00')
  /Producer (QuarkXPress(R) 8.16r2)
  /Title \<feff0063006d0079006b002d003200300030>

Removing the parentheses from the Creator and Producer values (or just the closing parentheses for that matter) allows the file to be validated successfully. According to my interpretation of the PDF reference (https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/pdf_reference_archives/PDFReference.pdf) the entries in the dictionary are just regular text strings and balanced parentheses should thus be allowed.

deanforsmith:
Tokenizer.java handles special characters through a series of if-else statements within the getNext method. [Line 248](https://github.com/openpreserve/jhove/blob/de1f5fd8427a0e8365b44ec095e34bf4d2cde8ad/jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/Tokenizer.java#L248) handles open parenthesis but there's no correlating check for close parenthesis. Suggest adding close parenthesis to this statement to handle both characters:
`else if (_ch == '(') {`
would become
`else if (_ch == '(' || _ch == ')') {`

david-russo:
[...]

PR #734, opened 2022-04-14
Move _rawbytes.add() to end of loop in processLiteral() so does not add close paren character

### LINKED ISSUE #696: Close parenthesis included in Literal _rawBytes, causes incorrect NameTreeNode.compareKey()
I *think* this is a bug, but it may also be badly labeled annotation links in a PDF. I'm logging it in case others find the same issue, or the underlying cause has additional implications. It is causing numerous errors like this: `edu.harvard.hul.ois.jhove.module.pdf.PdfInvalidException: Invalid indirect destination - referenced object 'bm_b11' cannot be found` in a particular set of PDFs. In this instance the problem seems to be that there is also a 'bm_b1' reference as well as 'bm_b11'. 

The problem stems from the position of this line in the Literal class:
https://github.com/openpreserve/jhove/blob/47f077fcb09d3dd5720aeeb4abf27de5dac48d19/jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/Literal.java#L164
A character is appended to `_rawBytes` regardless of whether it is a close parentheses that should end the Literal. That means the `_rawBytes` value ends with ",41" (close parentheses). 

Where this causes problems is when performing `NameTreeNode.compareKey()` here:
https://github.com/openpreserve/jhove/blob/47f077fcb09d3dd5720aeeb4abf27de5dac48d19/jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/NameTreeNode.java#L122
The compare process uses the `_rawBytes` - for comparison it basically truncates the longer key to a shorter key. If the two being compared are a different length, it will not reach the ",41"/ close paren of the longer `_rawBytes`, but it will be the last character compared on the end of the shorter one.  That means if the character that is compared to the close paren falls below 41, it returns -1 and exits the matching loop. This would be fine if the character was alpha-numeric, but in this case the character being compared is a null and so the annotation links, which are functioning when rendered, are causing validation errors in JHOVE. 

Example of the problem reference:
image
I think this should be valid?

To confirm this was causing the problem, I did a quick hack to move this line:
https://github.com/openpreserve/jhove/blob/47f077fcb09d3dd5720aeeb4abf27de5dac48d19/jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/Literal.java#L164
... to the last line of the for-loop since a close paren will cause a `return offset;` and the character will not be appended to the _rawBytes. Moving that line caused the error messages to stop.

karenhanson:
Resolves [issue #696](https://github.com/openpreserve/jhove/issues/696). 

In `Literal.processLiteral()` the close parenthesis character is included in the rawbytes sequence but not the string buffer. This causes a problem in `NameTreeNode.compareKey()` where shorter strings in UTF16 are incorrectly compared with longer strings that start the same way - because they match a null against a close parenthesis. Unless there is a reason to include close parenthesis in the _rawbytes, this appears to be a bug. If so, the fix is to move the `_rawbytes.add()` in `Literal.processLiteral()` to the end of the loop that builds the string so that the close parenthesis is not included in the _rawbytes. 

Noticed this update recently, so took a look: https://github.com/openpreserve/jhove/issues/358. Looks like these two fixes will work together since close parens that belong in the rawbytes will still be included.

### DIFF
--- a/jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/Literal.java
+++ b/jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/Literal.java
@@ -161,7 +161,6 @@ public class Literal
                 throw new EOFException (MessageConstants.PDF_HUL_10.getMessage()); // PDF-HUL-10
             }
             offset++;
-            _rawBytes.add (ch);
             if (_state == State.LITERAL) {
                 // We are still in a state of flux, determining the encoding
                 if (ch == FE) {
@@ -300,6 +299,7 @@ public class Literal
                     buffer.append ((char) utfch);
                 }
             }
+            _rawBytes.add (ch);
         }
     }
 


carlwilson:
Given the packaging issue will see if this can make the cut

Issue #640, opened 2020-08-26
Report a more informative error message for encrypted PDFs


bitsgalore:
Jhove's output could be more informative for PDFs that use any kind of encryption. Currently JHOVE reports the following error message in this case (tested with Jhove 1.24.1, 2020-03-16):

```
ErrorMessage: Compression method is invalid or unknown to JHOVE
``` 

This is not particularly helpful, especially because it provides no clue whatsoever about encryption being the culprit. Perhaps the error message could be changed to something more informative that at least mentions the encryption?

Example files:

- <https://github.com/openpreserve/format-corpus/raw/master/pdfCabinetOfHorrors/encryption_noprinting.pdf>
- <https://github.com/openpreserve/format-corpus/raw/master/pdfCabinetOfHorrors/encryption_nocopy.pdf>
- <https://github.com/openpreserve/format-corpus/raw/master/pdfCabinetOfHorrors/encryption_notextaccess.pdf>
- <https://github.com/openpreserve/format-corpus/raw/master/pdfCabinetOfHorrors/encryption_openpassword.pdf>

asciim0:
I strongly support this and can supply quite a few real-life examples.

BertrandCaron:
We recently ran into a similar problem, and our JHOVE users couldn't understand the problem because of the arbitrary message JHOVE returned. So I also support this issue.

bitsgalore:
I just stumbled across this issue again. In addition to the confusing message itself, the fact hat JHOVE exits with an error (instead of parsing the file) also doesn't seem quite the right behavior (except for PDFs with an "open password", which  cannot be parsed at all without a password). For all other cases (printing, copying and text access restrictions) other tools such as VeraPDF are perfectly capable of parsing PDFs with those restrictions.  And since the PDF spec allows these restrictions, they shouldn't result in a "Not well-formed" validation outcome.

bitsgalore:
Seems JHOVE 1.28 changes fixed this, so closing this issue now.

bitsgalore:
Re-opening as JHOVE still reports this error for some PDFs with security-related features. Examples:

- [e-book_79.pdf](https://wiki.opf-labs.org/download/attachments/101613571/e-book_79.pdf) ([jhove output](https://github.com/KBNLresearch/pdf-characterisation/blob/main/output/jhove-pdfhul-impact/pdf-hul-109/e-book_79-jhove.xml))

[...]

PR #746, opened 2022-05-03
Solution for some of "Compression method is invalid or unknown to JHOVE" errors as mention in #640

samalloing:
Hi @carlwilson 

Could you check the solution I created used in this pull request? This solves the problem of some encrypted files that give the following error message "Compression method is invalid or unknown to JHOVE". There is more information to extract from the encryption, but before I put some effort in looking into this, I want to be sure the solution I used is acceptable.

Thanks

Sam


carlwilson:
Nothing but white space changes here now @samalloing so it can be closed.

PR #749, opened 2022-05-27
Problem Filter as indirect object solved, fixes #672

### LINKED ISSUE #672: False positive for PDF-HUL-45 with Filter as indirect object
It seems that JHOVE throws the PDF-HUL-45 "Malformed FIlter" validation error when the filter is referenced as an indirect object.
Unfortunately I cannot share the file, but the example is as follows:
```
1402 0 obj
<</BitsPerComponent 8
/ColorSpace 1817 0 R
/Filter 1401 0 R
/Height 258
/Length 22703
/Subtype/Image
/Type/XObject
/Width 554>>stream
```
where the content of the referenced object 1401 is:
```
1401 0 obj
/DCTDecode
endobj
```
When I replace "/Filter 1401 0 R" with "/Filter /DCTDecode" and rebuild the xref, the PDF validates fine.

It appears that the according section in the code starts here:
https://github.com/openpreserve/jhove/blob/47f077fcb09d3dd5720aeeb4abf27de5dac48d19/jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/pdf/PdfStream.java#L154 

Maybe someone who can actually understand the code (not me) can verify this :-)

samalloing:
Hi @carlwilson ,

For this pull request I updated some constructors. Can you check if it is according to your standards?

Thanks

Sam


carlwilson (commented):
Hmmm, @samalloing, I like the fix but not the implementation. This isn't your problem. The issue is the division of responsibilities between the module, the parser and the PDFStream class. The module does quite a lot of parsing and already has a cyclic dependency with the parser, i.e. they both depend on each other. The module already knows about all manner of low-level PDF classes it shouldn't be concerned with. Introducing another cyclic dependency between stream and module is just wrong.

On the other hand, it's a good fix, and the cyclic dependency horse has well and truly bolted. Proper refactoring is what's required but that's more work than the fix. The fix with a cyclic dependency is a reminder that this is a pain point and needs addressing. Interested to hear your thoughts?

samalloing:
Hi @carlwilson 

Yes, I thought you wouldn't be happy with it :-) But I looked long for a solution and I hope the problem can be fixed. If you have a better solution that solves the problem that would be great of course.
But is there an issue to solve the cyclic dependencies in the module, so we can work on this and solve it gradually.

Sam

carlwilson:
As I said, it's not an issue of your making. If I'd seen an easier route I'd have gone for it. I'm coming round to the idea of merging this for now as it tracks a real dependency regardless of how ugly it is.

PR #750, opened 2022-05-27
TIFF: Continue validation even if data is unaligned

david-russo:
Storing data unaligned to word or byte boundaries can decrease read performance but needn't halt file validation entirely as it doesn't prevent the file from being read. This will allow other potentially more serious issues to also be reported.

This commit also makes the 'byteoffset' configuration option affect the reporting of unaligned IFDs the same way it does unaligned IFD Entry values (by treating them as valid when set to 'true').


carlwilson:
I've taken this one in in #840 thanks @david-russo

PR #763, opened 2022-06-23
DOC: Added Gzip info

charlottearmstrong246:
Added GIF and GZIP info
Please review before I add more rows

### DIFF
--- a/index.html
+++ b/index.html
@@ -53,7 +53,12 @@ title: JSTOR/Harvard Object Validation Environment
             <td>.gif</td>
             <td>image/gif</td>
           </tr>
-        </tbody>
+            <tr>
+            <td>GZIP-hul</td>
+            <td>Gzip</td>
+            <td>.gz</td>
+            <td>application/gzip</td>
+          </tr>        </tbody>
       </table>
     </p>
     <p>


index.html
```diff
@@ -53,7 +53,12 @@ <h2>Software</h2>
             <td>.gif</td>
             <td>image/gif</td>
           </tr>
-        </tbody>
+            <tr>
```
carlwilson: It's good to try to keep the indentation consistent, please. HTML is something of a nightmare to follow and tracing indents is your best bet for navigation.

index.html
```diff
@@ -53,7 +53,12 @@ <h2>Software</h2>
...
+            <td>GZIP-hul</td>
+            <td>Gzip</td>
+            <td>.gz</td>
+            <td>application/gzip</td>
+          </tr>        </tbody>
```
carlwilson: Again, it's good to keep unconnected `<tags>` on different lines with consistent indentation for readability.

carlwilson (changes requested):
Sorry to be fussy but it keeps things easy to read.

carlwilson (approved):

PR #777, opened 2022-07-21
Duplicate PDF-HUL-154 when one should be PDF-HUL-155

samalloing:
Hi @carlwilson 

in a previous commit an error message got messed up. This is a fix for that.

Sam


samalloing:
I’ll close this it accidentally includes another fix as well

PR #782, opened 2022-07-22
String property can be empty

samalloing:
Hi @carlwilson 

In PDF files the String Properties like Title can be empty. Now the value is null and will give a nullpointerexception when that happens. I updated the jHove Property, but I could also only update the addStringProperty in PDF module.

Sam


jhove-core/src/main/java/edu/harvard/hul/ois/jhove/Property.java
```diff
@@ -77,7 +77,12 @@ private void init (String name, PropertyType type, PropertyArity arity,
            types here may help avoid difficult chasing down
            of the bugs such mismatches cause. */
 	if (value == null) {
-	    throw new NullPointerException (CoreMessageConstants.EXC_PROP_VAL_NULL);
+		if (type == PropertyType.STRING) {
+			// a String value can be empty
+			value = "";
+		} else {
```
carlwilson, line 83: The else is unnecessary if it's not a string then we always want the exception which terminates.

carlwilson (changes requested):

jhove-core/src/main/java/edu/harvard/hul/ois/jhove/Property.java
```diff
@@ -77,7 +77,12 @@ private void init (String name, PropertyType type, PropertyArity arity,
            types here may help avoid difficult chasing down
            of the bugs such mismatches cause. */
 	if (value == null) {
-	    throw new NullPointerException (CoreMessageConstants.EXC_PROP_VAL_NULL);
+		if (type == PropertyType.STRING) {
+			// a String value can be empty
+			value = "";
```
carlwilson, line 82: Look at the [Codacy feedback](https://app.codacy.com/gh/openpreserve/jhove/pullRequest?prid=9847925). It might be a good idea not to reassign `value` and add an intermediary value instead. Parameter mutation is rarely a good thing ;)

carlwilson (changes requested):

jhove-core/src/main/java/edu/harvard/hul/ois/jhove/Property.java
samalloing: Hi @carl, can you look at the new proposal I created for this? https://github.com/openpreserve/jhove/commit/9181e61b44925b71871c9b744b7049f33a15c227. The code doesn't become better readable, but it does what I wanted to achieve with this pull request without reassigning. I tried different options, but there was always some reassignment happening.

PR #743, opened 2022-04-26
Unhandeld exception when Encrypt is a Dictionary and not an indirect …

samalloing:
…object
File used: https://github.com/mozilla/pdf.js/blob/master/test/pdfs/issue6010_1.pdf

PR #783, opened 2022-07-22
Encrypt can also be a Dictionary

samalloing:
Hi @carlwilson 

Encrypt can be also be a PdfDictionary (also used Arlington PDF model to verify)

A testfile: https://github.com/mozilla/pdf.js/blob/master/test/pdfs/issue6010_1.pdf

Sam

### DIFF
--- a/jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/PdfModule.java
+++ b/jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/PdfModule.java
@@ -53,7 +53,6 @@ import edu.harvard.hul.ois.jhove.Identifier;
 import edu.harvard.hul.ois.jhove.IdentifierType;
 import edu.harvard.hul.ois.jhove.InfoMessage;
 import edu.harvard.hul.ois.jhove.InternalSignature;
-import edu.harvard.hul.ois.jhove.Message;
 import edu.harvard.hul.ois.jhove.Module;
 import edu.harvard.hul.ois.jhove.ModuleBase;
 import edu.harvard.hul.ois.jhove.NisoImageMetadata;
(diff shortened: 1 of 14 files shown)


jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/PdfModule.java
```diff
@@ -1361,9 +1361,16 @@ protected boolean parseTrailer(RepInfo info, boolean prevOnly)
 				throw new PdfInvalidException(MessageConstants.PDF_HUL_75, // PDF-HUL-75
 						_parser.getOffset());
 			}
-			_encryptDictRef = (PdfIndirectObj) _trailerDict
-					.get(DICT_KEY_ENCRYPT); // This is at least v. 1.1
-			_encrypted = (_encryptDictRef != null);
+			PdfObject encryptObj =  _trailerDict.get(DICT_KEY_ENCRYPT);
```
carlwilson: This is a solid approach, but the if/else leaves the possibility that the instance returned here isn't "tested" at all. Put another way, what happens when `encryptObj` ISN'T an instance of `PdfIndirectObj` OR `PdfDictionary`? This may be an error that gets picked up elsewhere, but I wonder if there shouldn't be a default condition here that marks an error?

carlwilson (changes requested):

carlwilson:
@samalloing I've also created PR #803, which sets things up for this change. I've added you as a reviewer anyway, but the two want merging before trying to merge into integration. ~~I'm good to handle that but want to keep them separate until the review is resolved~~. I actually merged them up instead, I'll fix the tests once the changes are made.

BTW this appears to be a duplicate of #743 ? Please confirm and I'll close that PR.

carlwilson:
@samalloing I think I've fixed these issues in #810 where I've added the changes needed to report encryption in the new case. Finally, I've added a couple of examples to the corpora and patched the tests to work.

If you could review #810 we can get these merged up.

carlwilson:
Closed as contained in #810 now, thanks @samalloing

Issue #706, opened 2022-02-03
Allow the XML local schema configuration to specify URIs


pwinckles:
Currently, jhove allows users to specify local overrides for schema files through configuration params like:

```
schema=http://dublincore.org/schemas/xmls/simpledc20021212.xsd;/path/to/local/simpledc20021212.xsd
```

However, it would be nice to be able to specify a URL here too. For example:

```
schema=http://dublincore.org/schemas/xmls/simpledc20021212.xsd;http://digital.library.wisc.edu/1711.dl/XMLSchema-SimpleDC
```

Here is the implementation I made in our fork: https://github.com/UW-Madison-Library/jhove/compare/435f233...48d083b

PR #794, opened 2022-11-22
PATCH: Failing tests on GitHub Actions

carlwilson:
Some time in September the JHOVE XML module tests started to fail. This was caused by a 403 error when trying to load XSD schema used in the two XML test files in this PR. The schema references in the test files load external schema located at <http://www.loc.gov/standards/>. These are behind a Cloudflare proxy which is the source of the 403 for requests from GitHub Actions only. The tests run fine when run locally, locally using different JDKs via Docker, on GitHub actions machines locally and on Travis.

The solution has been to copy the schema to an OPF hosted domain for test purposes <https://schema.openpreservation.org/ois/xml/xsd/jhove/test/>. A robust implementation of schema checking in JHOVE is the real solution.


pwinckles:
For what it's worth, last year the dublin core schemas were also inaccessible due to cloudflare ddos mitigations. I resolved it by modifying jhove to allow specifying schema location overrides: https://github.com/openpreserve/jhove/issues/706

Issue #668, opened 2021-04-08
Unexpected exception java.lang.NullPointerException returned in validation result for Docusign summary PDFs 


ross-spencer:
Attached are two files exhibiting the same problem. They were created using the docusign demo: https://secure.docusign.com/demo and were created using its two export methods. 

For all intents and purposes it looks as if the validation completes, and the validation result is "Not well formed". This seems to be because of the error message raised: `ErrorMessage: Unexpected exception java.lang.NullPointerException`.

The results are below:

```plain
JhoveView (Rel. 1.25.0-SNAPSHOT, 2021-04-07)
 Date: 2021-04-08 09:51:15 CEST
 RepresentationInformation: /tmp/opf/combined_Please_review__sign_your_document.pdf
  ReportingModule: PDF-hul, Rel. 1.12.2 (2019-12-10)
  LastModified: 2021-04-08 09:05:38 CEST
  Size: 319340
  Format: PDF
  Version: 1.4
  Status: Not well-formed
  SignatureMatches:
   PDF-hul
  ErrorMessage: Unexpected exception java.lang.NullPointerException
   ID: PDF-HUL-94
  MIMEtype: application/pdf
  PDFMetadata: 
   Objects: 71
   FreeObjects: 3
   IncrementalUpdates: 2
   DocumentCatalog: 
    PageLayout: SinglePage
    PageMode: UseNone
   Info: 
    Title: 
    Author: 
    Subject: 
   ID: 0x35633136303966662d373265312d343133382d623963622d336132373233313264346263, 0xb71234e8ef9a42d94605e600888c58f8
   Filters: 
    FilterPipeline: FlateDecode
   Fonts: 
    Type0: 
     Font: 
      BaseFont: FDXESL+HelveticaNeue
      Encoding: Identity-H
      ToUnicode: true
    TrueType: 
     Font: 
      BaseFont: FDXESL+HelveticaNeue-Bold
      FontSubset: true
      FirstChar: 32
      LastChar: 87
      FontDescriptor: 
       FontName: FDXESL+HelveticaNeue-Bold
       Flags: 32
       FontBBox: -1018, -481, 1437, 1141
       FontFile2: true
      Encoding: WinAnsiEncoding
      ToUnicode: true
     Font: 
      BaseFont: FDXESL+HelveticaNeue
      FontSubset: true
      FirstChar: 32
      LastChar: 146
      FontDescriptor: 
       FontName: FDXESL+HelveticaNeue
       Flags: 32
[...]

PR #806, opened 2022-12-19
Initialize string-valued tokens with empty string

prettybits:
Related to #668, a null-pointer exception is thrown when adding string properties from docInfo in `readDocInfoDict` when the token values are an empty hex string. The parser only sets the value of a `Literal`'s underlying `StringValuedToken` when there are other tokens between `<` and `>`, so empty hex strings result in the token's `_value` being `null`, while `Property` expects an actual string.

This either needs to more stringent checking for `null` at the relevant call sites or just initializing the `_value` as an empty string for `StringValuedToken`. AFAICS this should pose no issues, does it? @carlwilson


carlwilson:
This doesn't create any issues and is a belt and braces catch for a fix submitted by @samalloing. I'm going to merge it all the same, the code base is too casual in its treatment of nulls.

prettybits:
Thanks for merging, I just saw there was also #820 that is effectively concerned with the same issue this PR also fixes. With this merged the added null check there in `addStringProperty` should actually not even be needed now, but for sure doesn't hurt. In any case, the affected PDFs validate all the same now. :)

PR #818, opened 2023-01-11
Make only one pass to xml-encode the values,

### LINKED ISSUE #817: [XML] Very long processing of XML output for big XMP content
When processing images (TIFF or JPEG) containing big amount (7Ko) of XMP data, the processing of the XML output takes more than 5 minutes !!!

Whe processing the same file but outputting in TEXT or JSON, the computing time is in seconds.

The encoding method for XML values goes through the values at least 5 times and uses a StringBuffer !!!

tledoux:
Goes down to 100ms instead of minutes.
Fixes #817


carlwilson (approved):
More efficient and more readable to boot, good stuff and thank @tledoux

PR #744, opened 2022-04-26
Unhandled exception when Size is not an integer

samalloing:
Example file: https://github.com/qpdf/qpdf/blob/main/qpdf/qtest/qpdf/bad10.pdf

PR #819, opened 2023-01-12
TEST: Added test file and fix to integration.

carlwilson:
Includes PDF size fix from @samalloing in #744


samalloing (commented):
This is ok

samalloing (commented):
this is ok as well

PR #820, opened 2023-01-13
FIX: Handling of PDF empty string properties

### LINKED ISSUE #809: False positive PDF-HUL-94 java.lang.NullPointerException
The attached file throws a PDF-HUL-94 NullPointerException but seems to be syntactially ok.
Cross-checked the file with Preflight, qpdf and pdfcpu. 
Pdfcpu did report an error for the /Tabs entry in the PagesDictionary as /Tabs is not supported with the claimed PDF version 1.4 of the file ... however, since JHOVE should only check agianst the 1.7 ISO spec, and is not version-feature-exact for anything y 1.7 this shouldn't make a difference,

Due to no futher info on the validation error and due to no negative results with other tools I consider this currently a false positive.

[algenreportNordsee_06082021.pdf](https://github.com/openpreserve/jhove/files/10331625/algenreportNordsee_06082021.pdf)

carlwilson:
- removed null check/fix in `Property` class which might have unintended side effects for other modules (it may be the right thing to do, but more thought and testing are required);
- removed the succession of null pointer catches in `PDFModule`;
- added specific handling of null String values in `PDFModule.addStringProperty` as this seems to apply the fix in the correct place with fewer potential side effects;
- added PDF regression test file for empty string properties; and
- copied affected integration test results files from 1.26.

This is a reworking of the contribution from @samalloing in PR #782 
Fixes #809


samalloing (approved):
Thanks for making it a clean solution again!

PR #840, opened 2023-03-16
TIFF: Continue validation even if data is unaligned

carlwilson:
Copy of #750 submitted by @david-russo, many thanks. Couldn't sort the conflict from the web UI.

Storing data unaligned to word or byte boundaries can decrease read performance but needn't halt file validation entirely as it doesn't prevent the file from being read. This will allow other potentially more serious issues to also be reported.

This commit also makes the 'byteoffset' configuration option affect the reporting of unaligned IFDs the same way it does unaligned IFD Entry values (by treating them as valid when set to 'true').


ross-spencer:
@carlwilson a small thing, may not be too important to David? but perhaps future commiters, if you add this to the commit message:  `Co-authored-by: NAME <co-author@example.com>` then it keeps that contribution with that author. The current tag only has the perception of adding someone because of GH's ability with markdown, but not Git's in-built collab features. More info [here](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors).

carlwilson:
Nice, and thank you very much @ross-spencer. That's obviously a better way of doing things :)

Issue #877, opened 2023-08-16
encodeValue and encodeContent do not escape some invalid XML characters 


karenhanson:
I came across an issue where a message contained invalid characters (specifically `0xfffe` and `0xffff`). This caused the XML output to be invalid when parsed... this is even though the messages are going through `Utils.encodeContent(String content)`, which should clean up the string for XML. [This documentation](https://en.wikipedia.org/wiki/Valid_characters_in_XML) indicates that the two characters are forbidden.

I will do a PR for a proposed fix to address this problem, but wanted to log an issue to attach the fix to. 

I also noted something else while troubleshooting:

The document linked above also lists "surrogates" as forbidden and says some characters are "discouraged though allowed." Of the discouraged characters, the JHOVE Utility only removes one... `0x7f`.  I thought there may be a standard function to clean XML that's usable and looked at `org.apache.commons.lang3.StringEscapeUtils.escapeXml10`. It handles surrogates, escapes the discouraged characters (which are XML version specific), but also encodes quote, greater than, and apostrophe in all cases... which may not be a good thing for readibility of messages. Anyway, I mention it because I think [the code](https://commons.apache.org/proper/commons-lang/javadocs/api-3.3/src-html/org/apache/commons/lang3/StringEscapeUtils.html#line.134) might be useful if it seems important to explore escaping further, or we need to handle the unicode surrogates.

PR #878, opened 2023-08-16
Removes invalid characters 0xfffe and 0xffff when cleaning XML

karenhanson:
This removes two additional forbidden characters when cleaning XML to be output by the XMLHandler in Utils.encodeContent() and Utils.encodeValue(). If you leave these characters in the XML, it generates an invalid character message during validation of JHOVE output.

Relates to https://github.com/openpreserve/jhove/issues/877


karenhanson:
This appears to address #880, which has a sample file attached that can be used for testing.

Update Oct 30, 23: from discussion on #880, this fixes the error message on that ticket, but does not address the underlying issue, which appears to be to do with how the bits are read in. The file on that ticket can be used to test though.

PR #893, opened 2023-11-28
Empty properties causing extra indents in XmlHandler

karenhanson:
Submitting this minor fix to move down the piece of code that increments the indent level in XmlHandler.showProperty. I kept noticing when running large batches the XML would gradually move across the page. I figured out that when Utils.isPropertyEmpty() comes back true, it increments the indent unnecessarily causing misalignment in the XML. This corrects the alignment issue.


karenhanson:
OK, I see it's failing integration tests because it's comparing to XML files with the previous indents. Not sure where to find those files to update them...

Issue #894, opened 2024-01-19
Missing local messages in modules lead to InvocationTargetExceptions


rsteph-de:
Running JHove in a Non-English environment I ran into lots of `InvocationTargetException`s if localized messages were missing.
Looking into the `ErrorMessage_de.properties` of different modules I discovered several issues with wrong / missing message properties.
This led to NULL values in the corresponding `MessageConstants` enum.  

1) As a quick fix you can already set a system property to force JHove to use the English messages.
``` java
System.setProperty("module.language", "en");
````
2) I am going to provide some corrections to the properties as Pull Request.

3) It would be good to have a fallback to the default message, if a message is missing for a certain language.
   This will usually happen when error messages are added or updated.

PR #895, opened 2024-01-23
provide some German messages (#894)

rsteph-de:
more in issue #894


carlwilson (approved):
Good work, nicely done. Thanks very much @rsteph-de

PR #897, opened 2024-02-02
Error when Image Height and Width is null

samalloing:
Hi @carlwilson ,

I created a small pull request to make parsing of Image Height and Width more robust. In the arlington model they are required attributes of XObject Image, so that's why it is an error (and not a warning)

Sam


carlwilson (approved):
Nice, thanks @samalloing

PR #899, opened 2024-02-16
Change in GUI MIX field name from mimetype to format name

### LINKED ISSUE #896: Different element name in UI and when exported as xml or txt - PDF module
hi,
I was investigating why PDFMetadata.Images.Image.NisoImageMetadata.MIMEType element in PDF module was not getting any values in Rosetta and found strange differences in xml and txt exports when compared to JHOVE UI.
JHOVE UI result looks like this 
image
MIMEType field with value of "image/jpg".

When exported as xml the (what i think is the same element) is labelled differently. See below:
image

Similarly in txt output:
image
I think in the xml/txt output is wrong label for MIMEType element. 
thank you
Jan
PDF attached but you can test on any PDF 
[PDFA_in_a_Nutshell_211.pdf](https://github.com/openpreserve/jhove/files/14031828/PDFA_in_a_Nutshell_211.pdf)

### LINKED ISSUE #898: bitsPerSampleUnit element not showing in UI Results in PDF module
hi,
it looks like that element of "bitsPerSampleUnit" is not showing in UI results, only bitsPerSampleValue (under repinfo.PDFmetadata.images.image.NisoImageMetadata.bitsPerSample). Tested in JHOVE 1.28, 1.26 and 1.24.
image

It shows up fine in xml export though.
image

samalloing:
Hi @carlwilson 

The GUI shows Mimetype for MIX image data, according to the specifications this should be Format Name. The export in txt and XML show the correct field, but the GUI of jHove showed the wrong field name. Fixes #896 

Sam

### DIFF
--- a/jhove-apps/src/main/java/edu/harvard/hul/ois/jhove/viewer/RepTreeRoot.java
+++ b/jhove-apps/src/main/java/edu/harvard/hul/ois/jhove/viewer/RepTreeRoot.java
@@ -889,7 +889,7 @@ public class RepTreeRoot extends DefaultMutableTreeNode {
 				"NisoImageMetadata", true);
 		String s = niso.getMimeType();
 		if (s != null) {
-			val.add(new DefaultMutableTreeNode("MIMEType: " + s, false));
+			val.add(new DefaultMutableTreeNode("FormatName: " + s, false));
 		}
 		s = niso.getByteOrder();
 		if (s != null) {


carlwilson (approved):

leninoc:
hi, thanks for including this fix. 
The fix covers PDF, PNG, GIF, JPEG, JPEG2000 and TIFF modules - I thought I would make it clear here, someone might need this detail later on, thanks!

PR #901, opened 2024-02-16
add BitPerSampleUnit to the GUI

### LINKED ISSUE #898: bitsPerSampleUnit element not showing in UI Results in PDF module

samalloing:
Hi @carlwilson 
The BitsPerSampleUnit is missing in the GUI. It is hard-coded as Integer, because this is how it is also done in the other handlers. According to the specification of MIX, it can also be float. but this is not supported by jhove.

fixes #898 

Sam


carlwilson (approved):
Thanks @samalloing approved and merged

PR #912, opened 2024-03-08
Update Dockerfile

cstollw:
jhove version increased


carlwilson:
Thanks for this, it's an obvious templating candidate going forward.

carlwilson:
I'm closing this as I've updated to the new release for now. I've opened #952 so that the versions in the POM and the Dockerfile stay in sync.

PR #942, opened 2024-08-15
FIX: PDF v2 date parsing

### LINKED ISSUE #923: PDF-hul: False positiv for invalid date --> update date format to accept PDF2.0 notation
PDF versions up to and including 1.7 expected date types to include a terminating apostrophe. With PDF 2.0 this is no longer required. Both versions should be accepted. Currently date values not containing the terminating apostrophe are reported as "Well-formed, but not valid" with "PDF-HUL-133 Improperly formed date" error.

Please upate the date type to allow both variations:
(D:YYYYMMDDHHmmSSOHH'mm)
and 
(D:YYYYMMDDHHmmSSOHH'mm')

where:
YYYY shall be the year
MM shall be the month (01–12)
DD shall be the day (01–31)
HH shall be the hour (00–23)
mm shall be the minute (00–59)
SS shall be the second (00–59)
O shall be the relationship of local time to Universal Time (UT), and shall be denoted by one of the characters PLUS SIGN (U+002B) (+), HYPHEN-MINUS (U+002D) (-), or LATIN CAPITAL LETTER Z (U+005A) (Z) (see below)
HH followed by APOSTROPHE (U+0027) (') shall be the absolute value of the offset from UT in hours (00–23)
mm shall be the absolute value of the offset from UT in minutes (00–59)
optionally followed by APOSTROPHE (U+0027)

carlwilson:
- PDF-hul now allows dates in the form `D:YYYYMMDDHHmmSSOHH'mm` as well as `D:YYYYMMDDHHmmSSOHH'mm'`;
- refactored date parsing routine to allow unit testing; and
- added unit tests for date formats.

Closes #923


jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/PdfModule.java
samalloing, line 1: Hi @carlwilson ,

With this commit PDF-HUL-133 is deprecated. Should this be documented?

Sam

carlwilson: Hi @samalloing , thanks for the look over, it's appreciated :)

In this case, I've actually chosen to throw the exception from the [`Literal.parseDate()`](https://github.com/openpreserve/jhove/pull/942/files?diff=unified&w=1#diff-55d1c7dc0d525504e504dbd8c2639f732c7285cbcc71c24b56198ecf537e6081L428) method instead of returning null, so it's not deprecated. I do now throw the exception in multiple places but there's the possibility of adding extra details as the reason can be pinned down. I hope this clears up your concerns :)

samalloing: Thanks for clarifying! That is much better indeed!

PR #965, opened 2024-09-17
fix castException with byte[]

### LINKED ISSUE #926: jHove Gui stops without output in WAVE module
Hi @carlwilson 
There is a casting error because in https://github.com/openpreserve/jhove/blob/3872bcec770491a5e947cb221a8a2cdb766dabfe/jhove-apps/src/main/java/edu/harvard/hul/ois/jhove/viewer/RepTreeRoot.java#L509 there is a Byte[] but the extraBytes in the Wave module module are byte[] (https://github.com/openpreserve/jhove/blob/3872bcec770491a5e947cb221a8a2cdb766dabfe/jhove-modules/wave-hul/src/main/java/edu/harvard/hul/ois/jhove/module/wave/FormatChunk.java#L201)

Sam

samalloing:
Hi @carlwilson 
In the Viewer there is a CastException in the Byte[] in some WAVE files

This fixes #926 

Sam

### DIFF
--- a/jhove-apps/src/main/java/edu/harvard/hul/ois/jhove/viewer/RepTreeRoot.java
+++ b/jhove-apps/src/main/java/edu/harvard/hul/ois/jhove/viewer/RepTreeRoot.java
@@ -485,8 +485,16 @@ public class RepTreeRoot extends DefaultMutableTreeNode {
                     break;
                 }
                 case BYTE: {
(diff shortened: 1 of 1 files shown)


carlwilson (approved):
I see the problem and that looks a straightforward fix. Thanks @samalloing

PR #997, opened 2025-02-06
Bugfix in JPEG2000 module

marhop:
The JPEG2000 module did not parse a JPEG2000 codestream correctly when the length of a tile-part in a SOT marker segment was set to 0. This means that the tile-part contains all data in the remaining code stream. (See https://web.archive.org/web/20130810200214/http://www.jpeg.org/public/fcd15444-1.pdf section A.4.2) JHOVE however would interpret this incorrectly as meaning "literally 0", thus expecting the beginning of the next marker segment where in fact the tile-part bytestream starts.


marhop:
Lots of example files that illustrate this bug here: https://www.opengeodata.nrw.de/produkte/geobasis/lusat/akt/dop/dop_jp2_f10/

Issue #1011, opened 2025-03-08
SLF4J: Class path contains multiple SLF4J bindings.


99brgs:
When using JHOVE classes inside of an application I developed, it cannot be executed because it says
SLF4J: Class path contains multiple SLF4J bindings.

After some research, I think that the issue comes from ext-modules and epubcheck library as it includes slf4j-noop that is not a best practice.
https://www.slf4j.org/codes.html#multiple_bindings

carlwilson:
Thanks for reporting this @99brgs, we will take a look and see if there's something that can be done for the next release.

coryz-fs:
I'm running into the same issue with the **jhove-ext-modules epubcheck** dependency.

`LoggerFactory is not a Logback LoggerContext but Logback is on the classpath. Either remove Logback or the competing implementation (class org.slf4j.helpers.NOPLoggerFactory loaded from jar:nested:/var/app/current/application.jar/!BOOT-INF/lib/jhove-ext-modules-1.32.1.jar!/)`

Normally I would just exclude epubcheck from my dependency since I'm not using the EpubModule, but it looks like that doesn't work since the jhove-ext-modules pom is using maven-assembly-plugin to include all the dependencies in the jar.

If you don't want to exclude slf4j-nop from your epubcheck dependency, could a standard jhove-ext-modules jar be provided in addition to the jar containing its dependencies so that we can choose to exclude it in our applications?

PR #1015, opened 2025-03-14
Upgrade EPUB module to use latest epubcheck 5.2.1

karenhanson:
Sorry I didn't get this in earlier, I know the RC just came out. It can wait for next release. 
This is a change to upgrade to latest epubcheck 5.2.1, which fixes a bunch of bugs. It includes adding back in the test to make sure the created date is populated. The created date was missing in the 5.1.0 release of epubcheck and has now been added back in. (https://github.com/w3c/epubcheck/issues/1532)

Side note: in the current JHOVE RC a warning now appears multiple times when you run the EPUB module - it's there before and after this PR. Not sure if this warrants further investigation:
```
Warning
   XSLT version ignored: only "3.0" is recognized
```
If it seems like something that should be addressed, I can look into it.

### DIFF
--- a/.github/workflows/test-pr.yml
+++ b/.github/workflows/test-pr.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   build:
     name: Checkout and Build
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
 
     strategy:
       matrix:
@@ -52,7 +52,7 @@ jobs:
 
   coverage:
     name: Quality Assurance
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     needs: [ build ]
 
     steps:


carlwilson:
Thanks for this @karenhanson, as you say it will have to wait for 1.36. I'm hoping that a dependency update might fix this: <https://github.com/openpreserve/jhove/issues/1011>

carlwilson:
Hi @karenhanson. Enough time has passed, and enough people are interested, that I'm looking to include this in the 1.34 release, which I'm currently putting together. With that in mind, I took a look at this warning:
```shell
Warning
   XSLT version ignored: only "3.0" is recognized
```
The phrase doesn't Google well, but that's because it's in the Saxonica Saxon-HE code which is in a zip file [here on GitHub](https://github.com/Saxonica/Saxon-HE/blob/main/11/source/saxon11-4source.zip). It's the configuration for that Saxon processor that's complaining:
```java
case FeatureCode.XSLT_VERSION:
    if (!"3.0".equals(value)) {
        makeErrorReporter().report(
                new XmlProcessingIncident("XSLT version ignored: only \"3.0\" is recognized").asWarning());
    }
    //getDefaultXsltCompilerInfo().setXsltVersion(v);
    break;
```
Given it's occurring that often, and there's little mention of XSLT in the epub source, I think it's the declaration of xslt2 query bindings in the schematron files for epub checker [which you can see here](https://github.com/search?q=repo%3Aw3c%2Fepubcheck%20xslt&type=code). I suspect that each time one of these is used, the warning is thrown.

A bit more digging and I can confirm that when [this line of the Epub checker code](https://github.com/w3c/epubcheck/blob/main/src/main/java/com/adobe/epubcheck/xml/XMLValidator.java#L205) is called with a Schematron file source then the warning is output twice. Naively switching the `queryBinding` attribute to `xslt3` does nothing to prevent the issue, but it doesn't appear to break anything either. The ePub checker CLI application doesn't show this behaviour. Whether this is due to a different execution path, or some kind of config that suppresses the message output, I'm not sure.

karenhanson:
Thank you for looking into the warning @carlwilson - do you think this is worth digging further or does it seem harmless enough to ignore for this version? Maybe there is a way to change the log level to supress warnings somewhere.

carlwilson:
We can proceed with using the PR as it stands. It doesn't break anything; in fact, it will address a security warning and update the `epubchecker` dependency. The warning text is output to `stderr` rather than `stdout`, so it can be filtered out by command-line usage:

```shell
jhove -m EPUB-ptc -hul -h xml -k ./accessible_epub_3.epub 2> error.txt
<?xml version="1.0" encoding="UTF-8"?>                                                         
<jhove xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://schema.openpreservation.org/ois/xml/ns/jhove" xsi:schemaLocation="http://schema.openpreservation.org/ois/xml/ns/jho
ve https://schema.openpreservation.org/ois/xml/xsd/jhove/1.9/jhove.xsd" name="Jhove" release="1.34.0-RC1" date="2025-06-26">
 <date>2025-06-27T08:58:19+01:00</date>                                                        
 <repInfo uri="./accessible_epub_3.epub">                                                      
  <reportingModule release="1.3" date="2023-06-12">EPUB-ptc</reportingModule>                  
  <created>2022-12-05T18:04:48+00:00</created>                                                                                                                                                
  <lastModified>2012-10-24T15:30:00+01:00</lastModified>                                                                                                                                      
  <size>4051892</size>                                                                                                                                                                        
  <format>EPUB</format>                                                                                                                                                                       
  <version>3.3</version>                                                                                                                                                                      
  <status>Well-Formed and valid</status>                                                                                                                                                      
  <mimeType>application/epub+zip</mimeType>                                                                                                                                                   
  <properties>                                                                                                                                                                                
...
 <checksums>
   <checksum type="CRC32">93d3a1e4</checksum>
   <checksum type="MD5">8465627fabc9ae16c790570de9137523</checksum>
   <checksum type="SHA-1">0856b0729fdefbfd3699461a6640cd8acd8f138d</checksum>
   <checksum type="SHA-256">67f75b8e3cd1abe4bb143d91d5424191d5af3115c9d26ff029a38e19f8d16feb</checksum>
  </checksums>
 </repInfo>
</jhove>

❯ cat error.txt
Warning 
   XSLT version ignored: only "3.0" is recognized
Warning 
   XSLT version ignored: only "3.0" is recognized
Warning 
   XSLT version ignored: only "3.0" is recognized
Warning 
   XSLT version ignored: only "3.0" is recognized
Warning 
   XSLT version ignored: only "3.0" is recognized
Warning 
   XSLT version ignored: only "3.0" is recognized
...
```

which will clean up the output for users. I tried a few mitigating tactics, updating the `epubchecker` project itself. As I noted above, changing the value of the `queryBinding` attribute does nothing to fix the issue. I also tried updating `saxon-HE` to 12.7, but that simply resulted in a more elaborate warning message:

```shell
Warning                                                                                        
  SXWN9020  XSLT version ignored: only "3.0" and "4.0" are recognized
```

I think that clearing up the output can be done by a logging config file somewhere, but that's a guess.

karenhanson (commented):
Looks good!

carlwilson:
Merged and closed with #1035

PR #1017, opened 2025-03-28
Markdown conversion of the gh-pages docs folder 

darrendignam:
I have changed most of the files from .html to .md and converted the contents to markdown.

I have added more id's to headings for internal page linking. 

I have added some visual quality of life improvements 

The new render can been seen live here:
https://jhove.darrendignam.com/


ross-spencer:
@darrendignam have you considered markdownlint for adding some extra consistency of the output here? 

Primary benefits are improved semantic markup, e.g. headings/code-block labels to identify language; layout; and a consistency that means the markdown can be processed in future by a larger number of markdown that aren't necessarily, expecting divergence from the spec. 

You can still see examples where GitHub once worked on more divergent Markdown, e.g. one of my [old repos](https://github.com/exponential-decay/checksum-comparator?tab=readme-ov-file#in-the-beginning) (look for the ## headings (Github flavor markdown used to accept no-spaces after the hash)).

Theoretically, it can promote better preservation like  anything looking at well-formedness and validity. 

I use this markdown lint tool for analysis in combination with pre-commit:

https://github.com/igorshubovych/markdownlint-cli

And I've seen folks have good results embedding "[prettier](https://prettier.io/blog/2017/11/07/1.8.0.html)" in CI workflows, and in pre-commit hooks but it's not my favorite tool to embed/configure. Benefit is that it is automatic.

Looking at results on the docs here, there are ~800 odd occurrences where the markdown might be tightened up. Including line-length (folks have different opinions on this one, another ~500 or so).

You may already be on it, but if not, maybe worth a look, and just drop me a message if I can help rig something up at all. It can always be done after this PR sometime down the line.

docs/about/index.md
```diff
@@ -0,0 +1,47 @@
...
+#### Become an OPF member
+
+By becoming a member of the OPF, organisations are directly helping to fund the hosting, maintenance, and development of the software, ensuring it is robust and reliable.
+
+Members can also choose to guide the roadmap and:
+- Sit on the JHOVE Product Board
+- Vote for bug fixes and enhancements to be included in each release cycle
+- Receive technical support
+
+Find out more: [http://openpreservation.org/about/join/](http://openpreservation.org/about/join/)
```
carlwilson, line 28: If the link text is just the link then it may be better to use `<link>` instead, see <https://www.markdownguide.org/basic-syntax/#urls-and-email-addresses>

docs/about/index.md
```diff
@@ -0,0 +1,47 @@
...
+<form action="https://www.paypal.com/cgi-bin/webscr" method="post" target="_top">
+  <input type="hidden" name="cmd" value="_s-xclick">
+  <input type="hidden" name="hosted_button_id" value="UKRJRX7A5ZJZL">
+  <input type="image" src="/img/paypal-donate.png" border="0" name="submit" alt="PayPal – The safer, easier way to pay online!">
+</form>
+
+#### Contribute to JHOVE
+
+We also welcome community contributions. We are looking for users and developers to join our active JHOVE community and help:
+- Test the software
```
carlwilson, line 43: Add a blank line before lists.

docs/documentation/dev-module/index.md
```diff
@@ -0,0 +1,749 @@
...
+---
+
+# Writing a JHOVE Module (*draft*, 2005-02-07)
+
+{: #module-interface .section-heading}
+## 1 The Module Interface
+
+All JHOVE modules implement the `module` interface. (Details of all interfaces and classes are available [here](/javadoc/).)
+
+{: .blockquote}
```
carlwilson, line 13: I'm not in love with the use of blockquotes here or the annotation used for them. I appreciate that you've replicated the original look and feel, and that's what I don't like.

- They're used inconsistently, at first for the code block but later for the explanations
- I'm not sure that nested blockquotes should be a thing
- The use of the class annotation should be replaced by the standard MD annotation `>` see <https://www.markdownguide.org/basic-syntax/#blockquotes-1>

docs/documentation/dev-module/index.md
```diff
@@ -0,0 +1,749 @@
...
+The `checkSignatures()` methods attempt to identify the object (represented as either a stream or random access file) using only internal signatures, i.e., magic numbers. Representation information about the object is returned through the `RepInfo` object.
+
+```
+public int parse (InputStream stream,    RepInfo info, int parseIndex) throws IOException;
+public int parse (RandomAccessFile file, RepInfo info) throws IOException;
+```
+
+> The parse() methods parse the object (represented by either a stream or random access file). Representation information about the object is returned through the RepInfo object. The stream version of parsemay be invoked multiple times, if it is necessary to do multiple passes on the data. On the first invocation of this method parseIndex is set to 0. If the method returns a non-zero value then it is invoked again, with parseIndex set to the return value.
+>
```
carlwilson, line 182: Blockquotes as they should be :)

docs/modules/ascii/index.md
```diff
@@ -0,0 +1,58 @@
+---
+title: ASCII-hul Module
+layout: page
+---
+
+# ASCII-hul Module
```
carlwilson, line 6: Perhaps we can generate some module details and add them here, e.g. current version, date and perhaps even the specifications covered.

carlwilson (changes requested):
A nice piece of work which needed doing. I will endorse @ross-spencer's comments regarding markdown lint. As well as running it and fixing the issues it might be good if we could incorporate the check as a QA gate via GH Actions.

On another note, while we are doing the above we could also add broken link checking as another QA gate.

I've added a few specific examples above to illustrate the above points. Let's have a chat about this and another pass, but good work overall :)

docs/documentation/dev-module/index.md
darrendignam: I was trying to get the converted documents to look as much like the originals as possible. In some cases using the markdown blockquotes was the most effective, but in others applying the blockquote css class was the alternative. In this case it was a code block that needed the style. In others it is because a table was required to be in a blockquote.

If we allow ourselves to deviate from the original look and feel of these docs, we can adhere to a stricter markdown syntax.

docs/about/index.md
darrendignam: I do like the idea of using a lint tool, as per @ross-spencer comments. I have run one and there are a lot issues right now. I will need to implement some custom rules as some of the issues are necessary for instance the github jekyll style decorators have to be before a block of text without an empty line to be effective. But I think adding a linter and correcting the current lint issues is a great first step, and then using github actions to maintain that level of QA is great. Cheers for the feeback

darrendignam:
I have had another go at this over at #1023
I have made a version that can pass a lint test with a bit of help. There are still some residual HTML elements that are hard to shake off. One option is to just not convert those files. But a better option would be to work to improve this conversion by updating the docs and allowing the look and feel, of say tables, or donation button links, so that we can convert to pure markdown, this conversion has tried to make the old and new look as close as possible to each other.

PR #1024, opened 2025-04-16
Fix developer extension version reporting

### LINKED ISSUE #890: Java Error on Jhove CLI by json formatter class java.lang.Double cannot be cast to class java.lang.String
If I run the cli with the -h option, the json output runs on the following error message.

SEVERE: class java.lang.Double cannot be cast to class java.lang.String (java.lang.Double and java.lang.String are in module java.base of loader 'bootstrap')
java.lang.ClassCastException: class java.lang.Double cannot be cast to class java.lang.String (java.lang.Double and java.lang.String are in module java.base of loader 'bootstrap')
at edu.harvard.hul.ois.jhove.handler.JsonHandler.showScalarProperty(JsonHandler.java:726)
at edu.harvard.hul.ois.jhove.handler.JsonHandler.showProperty(JsonHandler.java:658)
at edu.harvard.hul.ois.jhove.handler.JsonHandler.showListProperty(JsonHandler.java:790)
at edu.harvard.hul.ois.jhove.handler.JsonHandler.showProperty(JsonHandler.java:662)
at edu.harvard.hul.ois.jhove.handler.JsonHandler.showListProperty(JsonHandler.java:790)
at edu.harvard.hul.ois.jhove.handler.JsonHandler.showProperty(JsonHandler.java:662)
at edu.harvard.hul.ois.jhove.handler.JsonHandler.show(JsonHandler.java:439)
at edu.harvard.hul.ois.jhove.RepInfo.show(RepInfo.java:563)
at edu.harvard.hul.ois.jhove.JhoveBase.process(JhoveBase.java:649)
at edu.harvard.hul.ois.jhove.JhoveBase.process(JhoveBase.java:576)
at edu.harvard.hul.ois.jhove.JhoveBase.process(JhoveBase.java:576)
at edu.harvard.hul.ois.jhove.JhoveBase.process(JhoveBase.java:576)
at edu.harvard.hul.ois.jhove.JhoveBase.process(JhoveBase.java:576)
at edu.harvard.hul.ois.jhove.JhoveBase.dispatch(JhoveBase.java:479)
at edu.harvard.hul.ois.jhove.Jhove.main(Jhove.java:265)

XML output works, but looks a bit damaged:

 <values arity="Scalar" type="String">
           <value>&lt;x:xmpmeta xmlns:x="adobe:ns:meta/"&gt;
  &lt;rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"&gt;
    &lt;rdf:Description xmlns:pdfaid="http://www.aiim.org/pdfa/ns/id/" rdf:about=""&gt;
      &lt;pdfaid:part&gt;3&lt;/pdfaid:part&gt;
      &lt;pdfaid:conformance&gt;B&lt;/pdfaid:conformance&gt;
    &lt;/rdf:Description&gt;
    &lt;rdf:Description xmlns:dc="http://purl.org/dc/elements/1.1/" rdf:about=""&gt;
      &lt;dc:title&gt;
        &lt;rdf:Alt&gt;
          &lt;rdf:li lang="x-default"/&gt;
        &lt;/rdf:Alt&gt;
      &lt;/dc:title&gt;
    &lt;/rdf:Description&gt;
    &lt;rdf:Description xmlns:xmp="http://ns.adobe.com/xap/1.0/" rdf:about=""&gt;
      &lt;xmp:CreatorTool&gt;Konik Library&lt;/xmp:CreatorTool&gt;
      &lt;xmp:CreateDate&gt;2021-11-12T10:11:11+01:00&lt;/xmp:CreateDate&gt;
    &lt;/rdf:Description&gt;
    &lt;rdf:Description xmlns:pdf="http://ns.adobe.com/pdf/1.3/" rdf:about=""&gt;
      &lt;pdf:Producer&gt;Konik Library&lt;/pdf:Producer&gt;
    &lt;/rdf:Description&gt;
  &lt;/rdf:RDF&gt;
&lt;/x:xmpmeta&gt;</value>

prettybits:
Fixes #890

This fixes an exception that's surfaced when using the JSON handler with PDFs that list extensions in the document catalog. In this case, the parsed `Double` value of the extension's `BaseVersion` key was wrongly used for a `String` property. I also noticed that the version strings were parsed outside the `try`-block here, so the exception handling to report PDF-HUL-88 for extension's base versions didn't work as intended, this is also fixed.

While here, I noticed that extension base versions are checked whether they are the same as the file's PDF version to report an info message for PDF-HUL-87, equivalently to the reporting for mismatching document info version compared to the file header. Since Extension's `BaseVersion` "reflect the version of the standard that has been extended and not the version of this particular file" (see chapter 7.12.4 of the PDF specification) they need to be handled a bit differently. I aligned the check to not report anything for base versions lower than the PDF version as this is expected and only look for the case where the version is higher. Reporting PDF-HUL-87 for this case doesn't really fit then and would need a new error ID, I wasn't sure how to best go about introducing one. What do you think, @carlwilson?

### DIFF
--- a/jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/PdfModule.java
+++ b/jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/PdfModule.java
@@ -1749,9 +1749,10 @@ public class PdfModule extends ModuleBase {
                                     PdfSimpleObject BaseVersion = (PdfSimpleObject) extension.get(DICT_KEY_BASEVERSION);
                                     String infoVersString = _version;
                                     String versString = BaseVersion.getStringValue();
-                                    double ver = Double.parseDouble(versString);
-                                    double infoVer = Double.parseDouble(infoVersString);
                                     try {
+                                        double ver = Double.parseDouble(versString);
+                                        double infoVer = Double.parseDouble(infoVersString);
+
                                         // BaseVersion "shall be less than or equal to the PDF version" and "may be different from the version number in the document header or that supplied by the Version key in the Catalog [...] because it reflects the version of the standard that has been extended and not the version of this particular file"
                                         if (ver > infoVer) {
                                             // FIXME: this needs a separate error ID as this is a distinct case
--- a/jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/PdfModule.java
+++ b/jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/PdfModule.java
@@ -1760,7 +1760,7 @@ public class PdfModule extends ModuleBase {
                                                     MessageConstants.PDF_HUL_87.getId(), mess);
                                             info.setMessage(new InfoMessage(message));
                                         } else {
-                                            p = new Property(PROP_NAME_BASEVERSION, PropertyType.STRING, ver);
+                                            p = new Property(PROP_NAME_BASEVERSION, PropertyType.STRING, versString);
                                             _docCatalogList.add(p);
                                         }
                                     } catch (NumberFormatException e) {
--- a/jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/PdfModule.java
+++ b/jhove-modules/pdf-hul/src/main/java/edu/harvard/hul/ois/jhove/module/PdfModule.java
@@ -1752,7 +1752,9 @@ public class PdfModule extends ModuleBase {
                                     double ver = Double.parseDouble(versString);
                                     double infoVer = Double.parseDouble(infoVersString);
                                     try {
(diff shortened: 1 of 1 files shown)


carlwilson:
Thanks for this @prettybits. My thoughts are:
`PDF-HUL-87 = File header gives version as {0}, but catalog dictionary gives version as {1}` could possibly be worded better. Perhaps the Wiki is the place to improve the explanation. The case you're talking about is definitely a different case. I agree that a new error would be the best way of handling this. The steps are pretty straightforward:

- [x] read the specs and decide the nature of the test
- [x] implement test for condition
- [ ] decide on the wording of the message
- [ ] add message to properties files in PDF modules
- [ ] add message to MessageConstants loader in PDF module
- [ ] add message details to JHOVE wiki

That sounds like quite a bit once it's all done. Do you want to move this forward or should I look at then next tasks?

PR #1025, opened 2025-04-17
FIX: Ignore leading whitespace in hexadecimal strings

prettybits:
The PDF tokenizer already properly ignores whitespace characters when in the hexadecimal string state, but a check for whitespace immediately after the opening `<` was still missing.

This fixes PDF-HUL-11 False Positives we encountered in PDFs produced by various versions/forks of OpenOffice, where at least color tables of indexed color spaces were written as byte strings with a leading line feed character.


carlwilson:
Thanks for this @prettybits it's too late for 1.34 but I'll add it to be reviewed and merged for the next release.

carlwilson:
Nice and straightforward thanks, I'm updating and merging.

carlwilson:
This has changed the test output some, I'll patch the tests and PR the changes against this @prettybits

carlwilson:
The validation result for `test-root/corpora/errors/modules/PDF-hul/pdf-hul-29-govdocs-375118.pdf`. This previously was given the status `Not well-formed` with the message:

```xml
<message offset="4093" severity="error" id="PDF-HUL-11" infoLink="https://github.com/openpreserve/jhove/wiki/PDF-hul-Messages#pdf-hul-11">Invalid character in hex string</message>
```

It's now `Well-formed and valid`, the metadata indicates it's also produced by OpenOffice:

```xml
 <property>
  <name>Producer</name>
  <values arity="Scalar" type="String">
   <value>OpenOffice.org 2.3</value>
  </values>
 </property>
```

That seems in keeping with the PR, am about to PR a patch which will allow this to pass the integration checks.

prettybits:
Thanks, @carlwilson. As you say, the test file `pdf-hul-29-govdocs-375118.pdf` is precisely the kind of file expected to be found valid now when it previously wasn't. Thanks for opening a corresponding PR tending to the failing integration test!

PR #1070, opened 2025-07-31
Made the part of the byteoffset more clear

samalloing:
Added the error messages that want give an error anymore when byteoffset is set to true


carlwilson:
Thanks @samalloing straightforward doc change so merged.

PR #1085, opened 2025-10-21
Added detail to error message including entity URL in XmlModuleHandler

davidroncero:
I was struggling trying to identify which XSD was being rejected by a HTTP proxy. I wanted to configure XML-hul to map locally all the XSDs so jhove should not go outside to get them.

I added the `entityURL` in error message when `resolveEntity` was called.

Not a big change but it helped me to debug the problem in the final environment.


carlwilson (approved):
Thanks for this @davidroncero. It's nice and straightforward, and a sensible addition to the returned message. It'll be merged for the next release which should be this year.

PR #1097, opened 2026-04-22
Fix image links in docs

ross-spencer:
The references have changed following a previous commit.


ross-spencer:
NB. I'm reasonably sure these changes aren't making the CI fail, but not sure how to trigger a valid build. LMK if I need to change anything.

darrendignam:
Hi Ross

The CI fail is a java version conflict that we are aware of, and requires a proper overhaul of the build process that @carlwilson is currently looking at.

The images do work on the main docs site here:
https://jhove.openpreservation.org/getting-started/beginners/
But do not work on forked pages sites like this:
https://username.github.io/jhove/getting-started/beginners/

Because the base of the relative paths will make an image path will resolve to say
https://username.github.io/img/logo.png
but it needs to be:
https://username.github.io/jhove/img/logo.png

And while changing the paths to truly relative paths does address this issue, the preferred Jekyll approach is to use the relative_url filter:
https://gemdocs.org/gems/jekyll/4.2.2/Jekyll/Filters/URLFilters.html

And the syntax would look something like this:
{{ "/img/architecture.gif" | relative_url }}
And Jekyll will always create the correct full URL by add base.url to the URI taking into accoutn the current domain and sub folder location of the site.

We probably need to add an _config.yml - Jekyll will still work without this, but this is belts and braces.
docs/_config.yml

```
# Jekyll configuration for JHOVE documentation site

title: JHOVE
description: "JSTOR/Harvard Object Validation Environment"
url: "https://jhove.openpreservation.org"
baseurl: "" # Empty for custom domain at root

# Build settings
markdown: kramdown

# Exclude from processing
exclude:
  - Gemfile
  - Gemfile.lock
  - vendor

```

There are other images in the docs folder that your PR didn't address:
docs/_includes/header.html
docs/_includes/navbar.html

So we should address those too.

What do you think? 
@carlwilson uses a similar approach over on veraPDF - so his thoughts would be welcomed

ross-spencer:
@darrendignam thanks for the info.

I had a quick look and I just wanted to describe more the problem I thought I was solving here, which was enabling the images to render in markdown as well as in the built website, but, correct me if I am wrong, I think once we include `{{ ... }}` in the markdown it won't render in markdown, but links are probably more idiomatically written for Jekyll?

Looking around, I found this [Reddit post](https://www.reddit.com/r/Jekyll/comments/18dt5q4/comment/kcljo1s/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) which points here: https://github.com/JakeSteam/blog-personal/blob/1e3661331d9e844d5f808e06a9663e963bc0f6e5/_posts/2023-12-10-reviewing-every-jingle-jam-2023-game-2.md

And we can create a link that works quite nicely in Markdown and when built as follows:

```
[![JHOVE architecture overview](../../img/architecture.gif)](/img/architecture.gif) 
```

Can that work for this purpose? 

A benefit of having both images render in markdown and the site is we can point at both as resources, and it helps edit locally, but perhaps there is a way to do this in Liquid and markdown?

carlwilson:
Hi @ross-spencer. @darrendignam and I have been thinking about this one and we may have a better solution. My worry about relative links in this style is that they are fragile in a couple of cases:

- moving pages within the hierarchy will break the links; and
- the links won't work on GH pages sites published without a custom sub-domain, as this adds an extra path element at the start with the org name.

The issue might be better dealt with by abandoning a traditional web-site structure, where all images live under a single, top-level image directory. If the image only serves a single page then it could go in the same directory as the page, or an `img` sub-directory under that page. You could then use links like:

`[JHOVE architecture overview](architecture.gif)` or `[JHOVE architecture overview](img/architecture.gif)`. These would be robust in the cases mentioned above. Let me know your thoughts, thanks.

ross-spencer:
> If the image only serves a single page then it could go in the same directory as the page, or an img sub-directory under that page.

Sounds good to me @carlwilson! 

I can make a first attempt if it helps?

carlwilson:
Hi @ross-spencer if you fancy a go at it, that would be a help thanks. Much appreciated.

ross-spencer:
@carlwilson @darrendignam made what I think are all if not the majority of changes. The navbar logo is moved to the same depth as the favicons which means we remove the global img directory entirely. 

Let me know if you require any changes. Testing locally, and looking at pages via Github, e.g. https://github.com/openpreserve/jhove/blob/426af555b7c49cc44a19411ab6ce7a5a49141d5c/docs/documentation/architecture/index.md this looks like it works in Jekyll and GH Markdown.

PR #1106, opened 2026-06-13
add info about lower case URL for cached xsds in jhove.conf

asciim0:
added info about URL in XML-hul jhove.conf param needing to be in lower case


Kris-LIBIS:
See #1107 for a fix.

